PR #60, opened 2017-03-17
Try collapsing the what is information

BarrySmith:
Instead of making the user keep on jumping to other files and then back. They just click the > to open or collapse the details.

### DIFF
--- a/Site/Categories/Collaboration.md
+++ b/Site/Categories/Collaboration.md
@@ -1,57 +1,39 @@
 # Better Collaboration
 
-**Basic Description:**  Ways to facilitate and distribute work across teams, promote partnerships via software, and contribute to stronger communities.
+**Description:**  Ways to facilitate and distribute work across teams, promote partnerships via software, and contribute to stronger communities.
 
-**Brief Overview:** Software is the key crosscutting technology that sustains long-term CSE collaboration, encapsulating advances in applied mathematics, computer science, and domain-specific science and engineering for use by others.  We provide information for CSE projects of all sizes, ranging from small teams (for example, a professor and collaborating students) to broad partnerships that span different scientific, institutional, and geographic areas.  
-
-**Quote:** “Alone we can do so little; together we can do so much.”  — Helen Keller
+**Overview:** Software is the key crosscutting technology that sustains long-term CSE collaboration, encapsulating advances in applied mathematics, computer science, and domain-specific science and engineering for use by others.  We provide information for CSE projects of all sizes, ranging from small teams (for example, a professor and collaborating students) to broad partnerships that span different scientific, institutional, and geographic areas.  
 
 **Topics:**
 
-- **Version control:** Management of changes to files.  Also known as revision control, source control, or source management.
-
-  2-sentence description goes here. estimate of space is xxxxx xxxxx xxxx xxxx xxxx xxxxx xxxxx xxxx xxxx xxxx xxxxx xxxxx xxxx xxxx xxxx xxxxx xxxxx xxxx xxxx xxxx xxxxx xxxxx xxxx xxxx xxxx xxxxx xxxxx xxxx xxxx xxxx xxxxx xxxxx xxxx xxxx xxxx xxxxx xxxxx xxxx xxxx xxxx xxxxx xxxxx xxxx xxxx xxxx xxxxx xxxxx xxxx xxxx xxxx xxxxx xxxxx xxxx xxxx xxxx xxxxx xxxxx xxxx xxxx xxxx xxxxx xxxxx xxxx xxxx xxxx.
+- **Version control:** 
+Management of changes to files.  Also known as revision control, source control, or source management. xxxxx xxxxx xxxx xxxx xxxx xxxxx xxxxx xxxx xxxxx xxxxx xxxx xxx xxxxx xxxx.
 
     - [What is Version Control?](../CuratedContent/WhatIsVersionControl.md)
     - [How to Do Version Control with Git in Your CSE Project](..CuratedContent/HowToDoVersionControlWithGitInYourCseProject.md)
 
-- **Documentation:**  Written text or illustration that accompanies computer software to explain how it operates or how to use it.
-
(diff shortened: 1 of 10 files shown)


BarrySmith:
> On Mar 18, 2017, at 9:51 AM, Lois Curfman McInnes <notifications@github.com> wrote:
> 
> @curfman commented on this pull request.
> 
> Barry -- Thanks for thinking about user interface; this is very important, and we indeed appreciate suggestions about how to effectively present info (including levels of detail for topics) on the front-end site. The GitHub site (this page) is just where content contributors provide material that will eventually be displayed with an intuitive user interface on the front-end site (where most users will see the content). The front-end site is under development by our partners at Sandbox Studio and Parallactic. The format used in this file is input for a parser that then is used to insert content into the front-end site (which is built using Ruby on Rails).

   Ok

    Since HPCers generally do our own webpages we usually have very static pages with lots of links because doing anything else is too difficult. But this also means we "think" in terms of such pages.  I actually wanted to use an "in window" pop up (like you often see on commercial sites; they have the X in the upper right corner to remove them) for these small "blurbs" but it doesn't appear easy with markdown to do that. Presumably Sandbox Studio and Parallactic can do this easily.

    I am a bit concerned that if the contract with Sandbox Studio and Parallactic lapses in the future for any reason we may not be able to maintain the webpages they have made?

  Barry

curfman:
As part of the contract for phase-1 work on the site (current scope), Sandbox Studio and Parallactic will provide training and instructions so that we can maintain the site.   We may work with them to extend functionality (in a phase-2 project) as determined by feedback from site users.

markcmiller86:
FWIW, I'll second Barry's idea of expandable/collapsable blurbs underneath major headings instead of sep. pages. I think I understand Lois' explanation that some 3rd party web experts are going to take "raw inputs" from Github to "make the real hosted site". However, I also share Barry's concern about long-term viability of that approach. OTOH, with combination of github, Jekyll w/Liquid tags and SASS, I dunno why we can't own the whole site ourselves? I have worked in modalities where we have sep. web experts doing the web work and I have always found them frustratingly slow, unsatisfying in final result and not easily changed once created.

FWIW, I don't think we're doing stuff here that requires a CMS even for these pages. Our content isn't like a store always trying to look new and fresh. CMS are nice but they're often way overkill for what I need.

BarrySmith:
> OTOH, with combination of github, Jekyll w/Liquid tags and SASS, 

wow, Mark, with your experience sounds like we should hire you to build our web pages :-).

Since you know so much, the concern I have with my "collapse" model (and what I could find about open source pop up models on the web) is that the text that goes into the "collapse" must be stored in the same file as the outer page. This is a terrible data model because then there is no reasonable way to display that same text in other locations. Hence my concern that the "open source" non-commercial web display tools don't have flexible "data models" that I think we want to use.

curfman:
Hi Mark and Barry -- Thanks for your comments; agreed that expandable/collapsable blurbs are a good idea, and we are exploring for the front-end.   The GitHub site provides content for the front-end, so it is effectively serving as a content management system.    We'd be glad to discuss more details about approach and long-term strategy ... will put that on the agenda for sometime when we can talk rather than just writing.  Keep those good ideas coming :-)  Thanks.

markcmiller86:
On Mon, Mar 20, 2017 at 7:49 PM, Barry Smith <notifications@github.com>
wrote:

> OTOH, with combination of github, Jekyll w/Liquid tags and SASS,
>
> wow, Mark, with your experience sounds like we should hire you to build
> our web pages :-).
>
Well, I claim only enough knowledge of those technologies to conclude its
possible.

I have been desiging a couple of sites using Jekyll, SASS and Liquid and it
takes a minute to get oriented but then its pretty cool. Basically, its a
code-generation approach to a mostly-static site construction. So, it
replaces a CMS with a code-generation style of site creation. But, its
tightly integrated with github too so that as you commit, github will
re-gen the site. The technology was originally developed to support blog
sites but its been extended to do a lot more. For example, this site is
sourced on github here...

https://github.com/home-assistant/home-assistant.github.io

and served from github here...

https://home-assistant.io


> Since you know so much, the concern I have with my "collapse" model (and
> what I could find about open source pop up models on the web) is that the
> text that goes into the "collapse" must be stored in the same file as the
> outer page. This is a terrible data model because then there is no
> reasonable way to display that same text in other locations. Hence my
> concern that the "open source" non-commercial web display tools don't have
> flexible "data models" that I think we want to use.
>
I think the code generation approach can actually support the kind of "data
model" you are thinking where the text blurbs live each in their own file
and then you control, during site generation, how that content gets linked
into the rest of the site. One way, its collapse/expand content. Another
way, its in-line, indented content. Still another way, its follow-this-link
content.

All that said, I don't wanna become a site developer ;) At least not by
myself. I think this Jekyll stuff is straightforward enough, most people on
the team could pick it up. Heck, its just some conventional folder
hierarchies + content snipits + Liquid coding and we're all coders ;)

BarrySmith:
> On Mar 23, 2017, at 11:33 PM, Mark C. Miller <notifications@github.com> wrote:
> 
> On Mon, Mar 20, 2017 at 7:49 PM, Barry Smith <notifications@github.com>
> wrote:
> 
> > OTOH, with combination of github, Jekyll w/Liquid tags and SASS,
> >
> > wow, Mark, with your experience sounds like we should hire you to build
> > our web pages :-).
> >
> Well, I claim only enough knowledge of those technologies to conclude its
> possible.
> 
> I have been desiging a couple of sites using Jekyll, SASS and Liquid and it
> takes a minute to get oriented but then its pretty cool. Basically, its a
> code-generation approach to a mostly-static site construction. So, it
> replaces a CMS with a code-generation style of site creation. But, its
> tightly integrated with github too so that as you commit, github will
> re-gen the site. The technology was originally developed to support blog
> sites but its been extended to do a lot more. For example, this site is
> sourced on github here...
> 
> https://github.com/home-assistant/home-assistant.github.io
> 
> and served from github here...
> 
> https://home-assistant.io

  Thanks. The site is pretty but has no examples of "collapse" or pop up, the pages people see are static (nothing to steal from here).

markcmiller86:
Here's an example of a Jekyll created page with collapse...

http://idratherbewriting.com/documentation-theme-jekyll/mydoc_faq_layout.html

tgamblin:
Software carpentry's lessons use nested collapsible content for the exercises and solutions.  See the [git-novice](https://swcarpentry.github.io/git-novice/09-conflict/) lesson and its [github repo](https://github.com/swcarpentry/git-novice).  Also note that they're hosted on github.io.

BarrySmith:
> On Mar 24, 2017, at 10:18 AM, Mark C. Miller <notifications@github.com> wrote:
> 
> Here's an example of a Jekyll created page with collapse...
> 
> http://idratherbewriting.com/documentation-theme-jekyll/mydoc_faq_layout.html

 Nice.

maherou:
We have talked about the possibility of eventually hosting the BSSW.io site on GitHub, but I don't think we are prepared right now to take on that task and at the same time produce a polished experience.

The ruby-on-rails frontend should work well for us for some time, maybe even indefinitely.

maherou:
Committing this version of the file.  We can revisit dynamic formats in the future.

Issue #248, opened 2017-12-11
Article idea: Experience with adding unit testing to existing C++ project


markdewing:
A description of our experience adding unit testing using Catch (now Catch2) to an existing C++ code (QMCPACK)

Some possible points to cover:
1. Why Catch was chosen over other unit test frameworks
2. Integrating Catch with the project's CMake build system
3. Some small issues with Catch and scientific software - MPI, complex numbers

maherou:
Mark,

This looks like a very nice contribution.  We would be happy to receive it.  If you want to work on a draft and submit it as a pull request, we are happy to take a look and give feedback if needed.

Thank you for your interest.

rinkug:
This is done: https://bssw.io/items/unit-testing-c-with-catch

PR #284, opened 2018-06-26
Initial draft for unit testing with Catch.

markdewing:
From issue #248


rinkug:
This is done: https://bssw.io/items/unit-testing-c-with-catch

PR #291, opened 2018-08-28
Update ScienceAndSocialMedia.md

tscheibe:


pagrubel:
Nice article. I have a personal twitter account that I don't even use. I may have even deleted it. This makes me thing I should open up a professional one.

PR #313, opened 2018-11-01
Reduce to single link

shuds13:


maherou:
@clararaubertas: Can you check why this PR is failing?  May be a non-issue.

clararaubertas:
Looks like this is related to the python script syntax check on this end, which I haven't been involved with.

maherou:
@william76, Can you check into this?

PR #316, opened 2018-12-20
Rearrange article as more of an experience report

markdewing:
Changes based on email conversation with David Bernholdt.


curfman:
Thanks for the edits, Mark; very nice.  Next step is a pass by our technical editor; then we can release on the site and highlight in the January BSSw Digest.

PR #321, opened 2019-02-09
Apollo Guidance Computer Part 1

markcmiller86:
Here is part 1 of the Apollo Guidance Computer.

There is an associated pull request on the images repo, https://github.com/betterscientificsoftware/images/pull/2

After the above PR is merged, the markdown document here will need to be updated to point to the images on the *master* branch instead of the *blog_agc_part1* branch (e.g. replace string ``blog_agc_part1`` everywhere with ``master``)

All images are either NASA public domain or I've recieved permission from Raytheon Public Affairs.

Wish I could use [this image](https://picclick.com/Apollo-11-CSM-LM-Docked-Over-Moon-382266726290.html) for the spacecraft.


bernhold:
Merged so that we can preview.

PR #322, opened 2019-02-13
Craftsman blog post

prwolfe:
Initial blog post as requested by Lois. This still has the default image which at the very least needs replaced. I am opening an e-mail conversation about that with Matt.


maherou:
I really like this article.  I think it will be a nice addition.

prwolfe:
Thanks Mike! Any advice on getting the Review and Approval done in time?

Issue #331, opened 2019-03-28
Add new curated content for "Working Effectively with Legacy Code"


bartlettroscoe:
As per the approved proposal, I would like to submit a curate content article for the excellent "Working Effectively With Legacy Code" book.

bartlettroscoe:
@curfman, PR #359 is the completion of the "Working Effectively Legacy Code".

bernhold:
👍

PR #332, opened 2019-03-28
Initial content for WorkingEffectivelyWithLegacyCode (#331)

bartlettroscoe:
CC: @curfman, @markcmiller86 

Curated article for #331.


bartlettroscoe:
NOTE: I create this branch and PR completely on the GitHub website to show you can do it.  I will document the exact process soon.

CuratedContent/WorkingEffectivelyWithLegacyCode.md
```diff
@@ -0,0 +1,34 @@
+# Working Effectively with Legacy Code
+
+## Feathers, Michael, 2005
+
+This is really the book that gives one hope that we can take very bad software
```
markcmiller86: Not sure we wanna use the word "bad" here. Maybe a phrase "software written with obsolete development processes"

bartlettroscoe: Yes, how about "messy" or "hard to work with" or "hard to maintain"?

bartlettroscoe: @markcmiller86, please take a look at my latest commit 4b0fca2 and see if this is better.

Also note that I switched to the one-line-per-sentence format which massively improves concurrent editing and merging.

markcmiller86: I think any of those are better. You could get cute and coin a new phrase like "fossil-ware" or "vintage-ware" "bygone-ware" ;)

markcmiller86:
I wanna add a really inflammatory comment here just to see if/how we can hide it later.

So, consider this an inflammatory comment ;)

markcmiller86:
Ok, so I learned this...

> People with write access to a repository can hide, unhide, edit, or delete comments on issues, pull requests, and commits. 

To hide (or delete) a comment, go to the elipsis (three dots) in the upper right corner and pull down to either `Hide` or `Delete`. When you `Hide`, GitHub gives you a multiple choice set of reasons for  hiding the comment including `Off Topic`, `Spam`, `Abuse`, `Obsolete`, etc. I wonder if choosing `Abuse` might count negative against someone's GitHub brownie points? I dunno

> Hidden comments are minimized but people with read access to the repository can expand them.

I honestly don't know what this means. In theory, anyone with a browser can go to GitHub and look at a repository so doesn't the above mean *anyone* can expand the comment in which case what does *hiding* really achieve? I will have to test by logging out

markcmiller86:
Ok, so when I sign out of GitHub, I am not able to expand a hidden comment. So, I think this functionality allows is to mitigate any of these circumstances in the unlikely event that they come up.

pagrubel:
Did you check if anyone not a member of the repo that is signed in can see it?

Pat

markcmiller86:
I did not because I am not "not a member" of the repo and so not easily possible for me to test.

That said, my presumption is that if it is hidden when I am signed out, it is also hidden to non-members else, it would be visible to all on GitHub and that isn't very descriminating ;)

Mark


"Patricia Grubel" wrote:

Did you check if anyone not a member of the repo that is signed in can see it?

bartlettroscoe:
@prwolfe , after having merged your suggested changes (which were veyley good), do you approve this PR?

CuratedContent/WorkingEffectivelyWithLegacyCode.md
```diff
@@ -0,0 +1,25 @@
...
+
+#### Contributed by [Roscoe A. Bartlett](https://github.com/bartlettroscoe)
+
+#### Publication date:  March 28, 2019
+
+<!---
+Publish: yes
+Categories: development, reliability, skills
+Topics: refactoring, design, software engineering, testing, Personal productivity and sustainability
+Tags: book
```
prwolfe: Perhaps the isbn could go here?

CuratedContent/WorkingEffectivelyWithLegacyCode.md
```diff
@@ -0,0 +1,25 @@
+# Working Effectively with Legacy Code
+
+## Feathers, Michael, 2005
```
prwolfe: Need isbn

bartlettroscoe: @prwolfe,

Looks like if we add the ISBN to this document, it will be the first:

```
$ cd betterscientificsoftware.github.io
$ find . -type f -exec grep -nHi isbn {} \;
[empty]
```

Therefore, I don't know where to put it.  But I will put it somewhere.

bartlettroscoe:
@prwolfe said:

> We do need the isbn otherwise it's good to go.

I added ISBN and an image.

CuratedContent/WorkingEffectivelyWithLegacyCode.md
prwolfe: Interesting - we should ask if there is a reason not to. It just seemed like the right thing to my mind.

bartlettroscoe:
After talking with a Sandian last week, it occurred to me that many people in this community are really not aware of the advice in this book.  While I added a very short outline of the of "Legacy Software Change Algorithm", I thought it might be good to put in more of the outline material in:

* http://trac.trilinos.org/wiki/TribitsLegacySoftwareChangeAlgorithm

What all of the content on that short page be too long for  currated content article?  I just fear that people will not read the book and if they see this outline content, perhaps that is enough to start a conversation with people?

pagrubel:
Is there anyway you could consolidate what's on that page. I think its a bit long for a curated article, but your current article is short. Do we have a limit?

bartlettroscoe:
@pagrubel said

> Is there anyway you could consolidate what's on that page?

Yes.

> I think its a bit long for a curated article, but your current article is short. Do we have a limit?

That is my question.  What length would be considered too long?

This is a **really** important book and important material for CSE so getting the basics of this approach into people's brains is really desirable, IMHO.

rinkug:
This PR was extended in a new PR #359

PR #344, opened 2019-04-23
Added SAND#

ghammond86:
Sorry.  Another pull request to add the SAND #.


curfman:
@william76 -- Do you have insight into why this test failed?  The article looks fine to me.

william76:
not sure, looks like something caused an unhandled error in the script.  
I'm in Albuquerque attending the Kokkos Usersgroup meeting today, I'll have a look at the script and see if I can figure out what's going on in a bit. =)

william76:
@curfman 
I took a quick look at the metadata on this article and found a couple issues.

The meatadata is:
```
<!---
Publish: preview
Categories: reliability, development
Topics: testing, continuous integration testing, documentation
Tags: bssw-blog-article
Level: 2
Prerequisites: default
Aggregate: none

SAND #: SAND2019-4553 S
Image owned or licensed by Sandia
--->
```

The traceback on the checker came from the empty line in the metadata.  That's a bug and I'll get that fixed on the checker script.

The other thing that the script didn't hit but would also cause a problem in the metadata is the line:
```
Image owned or licensed by Sandia
```
This doesn't fit into the `key: value` formatting that we expect to see in the metadata section by the checker...  We should discuss whether or not freeform text should be allowed in our metadata.  If we do this, we should put some constraints in place or add a new catch-all rule such as free form text lines should have a `#` on column 1 or a `:` or something so I can write the checker to properly ignore those lines rather than trying to make sure it's a properly formatted metadata entry.

PR #358, opened 2019-05-17
a bunch of potential changes

danielskatz:
I think we should also list the talks themselves briefly


bernhold:
Thanks @danielskatz, improvements all.  Undecided about listing the talks. Maybe *instead* of the abstracts? That would keep the length in check.

Issue #360, opened 2019-05-20
"When is it better not to write automated tests?" Blog article


bartlettroscoe:
CC: @curfman, @maherou, @bernhold 

I have written a draft for a short blog article "When is it better not to write automated tests?".  I will have to send it through SNL R&A before I can post it here but I can send a draft to people offline in email to take a look at first.

I feel this goes with #331 "Working Effectively with Legacy Code" very nicely.

If interested, I will send out a draft of the formatted article.

markcmiller86:
I wonder if it makes sense to shorten #331 quite a bit and use the tandem blog article to capture what you do not include in #331 and, of course, with a link from the curated content article to this tandem blog article.

bartlettroscoe:
In my opinion, a blog article should have mostly unique material.  The material in this PR branch is really just a very short summary of existing material in the book "Working Effectively with Legacy Code".  Think of it as an extremely short Cliff Notes for "Working Effectively with Legacy Code".  Can there be some variability in curated content to allow this?

curfman:
In my view, this material would be a good fit for either curated content or an article (distinct from a blog article).  The boundary between the 2 is somewhat fuzzy (lengthier content generally in articles, though to date there have been just a few); the resource would appear the same on the frontend regardless (as a 'resource') rather than blog.

bartlettroscoe:
FYI: I submitted a draft to the SNL R&A process and I sent a draft to @curfman, @bernhold, @maherou, @elaineraybourn, and @markcmiller86 to get some preliminary feedback.  Once this is approved by SNL for release, I will post a PR if there is interest.

bartlettroscoe:
Comment from @curfman below.

Once I post the PR, I can add some section headings to break this up some to make it easier to browse.

---

Hi Ross – Thanks for proposing this article and writing a complete draft.  I believe this would be a very nice BSSw blog article as a complement to your other resource about the Feathers book.  I am sure that others who are considering strategies for expanded testing would benefit from considering these issues.

From a stylistic point of view, I recommend inserting a few subheadings to guide the reader.  We could schedule this to appear in late June or July.

Others may have additional comments about content.  

Best,
Lois

rinkug:
[...]

PR #366, opened 2019-05-25
Add requested features to wikize_refs.py

markcmiller86:
This update does the following
- Modularizes the code quite a bit
- Stores up output file lines in a list before writing it
- Makes output file read-only by default
- Adds option to disable read-only
- Adds option to specify output file name
- Change name from wikize-refs.py (which `import wikize-refs` fails due to dash (`-`) char) to wikize_refs.py
- Make all error exit calls pass `1` as in `exit(1)`

### DIFF
--- a/Articles/Blog/wikize_refs.py
+++ b/Articles/Blog/wikize_refs.py
@@ -42,7 +42,7 @@ To process a file...
 """
 
 from optparse import OptionParser
-import re, os
+import re, os, stat, tempfile
 
 def parse_args():
     """
@@ -52,31 +52,32 @@ def parse_args():
 
     parser.add_option("--warn",
                       default=False,
-                      dest="warn",
                       action="store_true",
                       help="[Optional] Warn instead of error during sanity checks.")
+
+    parser.add_option("--no-rdonly",
+                      default=False,
+                      action="store_true",
+                      help="Disable making output file read-only.")
+
+    parser.add_option("-o", "--outfile",
+                      help="Specify output file name instead of ")
 
     opts, mdfiles = parser.parse_args()
     return opts, mdfiles
 
 #
-# Read the command line
-#
-opts, mdfiles = parse_args()
-vopts = vars(opts)
-
-#
 # Process the original md file grepping for
 # footnotes and reference definitions
 #
-fn_handles = set()
-other_lines = []
-comment_lines = []
-original_refs = []
-ref_map = {}
-in_comment = False
-for mdfile in mdfiles:
-    with open(mdfile, 'r') as mdf:
+def process_input_file(filename):
+    fn_handles = set()
+    other_lines = []
+    comment_lines = []
+    original_refs = []
+    ref_map = {}
+    in_comment = False
+    with open(filename, 'r') as mdf:
         for mdfl in mdf.readlines():
             # grep for ^[xxx]: URL "Short Description {Formal Bibliographic data}"$
             mdfparts = re.search("^\[([a-zA-Z0-9_-]*)\]: (.*) \"(.*) {(.*)}\"$", mdfl)
@@ -108,37 +109,38 @@ for mdfile in mdfiles:
                 for fn in fns3: fn_handles = fn_handles.union(set(fn))
                 other_lines += [mdfl]
 
+    return fn_handles, other_lines, comment_lines, original_refs, ref_map
+
 #
 # Sanity checks for footnote references and the reference list
 #    - ensure every footnote references an existing item in the ref list
 #    - ensure every ref list item is referenced at least once
 #
-ref_handles = set(ref_map.keys())
-missing_refs = fn_handles - ref_handles
-if missing_refs:
-    print "Some footnotes never appear in the references..."
-    print "   ", [x for x in missing_refs]
-    if not vopts['warn']:
-        print "Correct above issues and re-try..."
-        exit()
+def sanity_checks(fn_handls, ref_map, warn):
+    ref_handles = set(ref_map.keys())
+    missing_refs = fn_handles - ref_handles
+    if missing_refs:
+        print "Some footnotes never appear in the references..."
+        print "   ", [x for x in missing_refs]
+        if not warn:
+            print "Correct above issues and re-try..."
+            exit()
 
-missing_fns = ref_handles - fn_handles
-if missing_fns:
-    print "Some references never appear in a footnote..."
-    print "   ", [x for x in missing_fns]
-    if not vopts['warn']:
-        print "Correct above issues and re-try..."
-        exit()
+    missing_fns = ref_handles - fn_handles
+    if missing_fns:
+        print "Some references never appear in a footnote..."
+        print "   ", [x for x in missing_fns]
+        if not warn:
+            print "Correct above issues and re-try..."
+            exit()
 
-#
-# Write the re-numbered and re-formatted md file
-# 
-with open("%s-wikized.md"%os.path.splitext(mdfile)[0], 'w') as wmdf:
+def generate_output_file_lines(other_lines, original_refs, ref_map, comment_lines):
+    outlines = []
 
     #
     # write warning comment about this being auto-generated
     #
-    wmdf.write("<!--- WARNING: Auto-generated with wikize-refs.py from %s --->\n"%mdfile)
+    outlines.append("<!--- WARNING: Auto-generated with wikize-refs.py from %s --->\n"%mdfile)
 
     #
     # Write all non-reference definition lines (e.g. article content) first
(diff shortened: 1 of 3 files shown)


Articles/Blog/wikize_refs.py
```diff
@@ -0,0 +1,265 @@
...
+        if not no_rdonly:
+            os.chmod(out_filename, stat.S_IRUSR | stat.S_IRGRP | stat.S_IROTH)
+
+def main():
+
+    #
+    # Process command line
+    #
+    opts, mdfile = parse_args()
+    vopts = vars(opts)
```
bartlettroscoe: Never saw this usage before.  I always just gave the option an explicit name in the `parser.add_option()` command.  That makes it explicit what the option should be called in Python code and makes it easy to use.

Articles/Blog/wikize_refs.py
```diff
@@ -0,0 +1,265 @@
...
+given the name infile.md and is made read-only. However, you can set
+the name of the output file instead and disable the read-only setting.
+By default, failed sanity checks cause an abort. However, this can
+be disabled by setting them to warn.
+
+To process a file...
+
+    ./wikize_refs.py foo.md
+
+...moves foo.md to foo.src.md and the new file is foo.md
```
bartlettroscoe: I think this is confusing to be moving file names around like this.  I would not rename any files.  I would just make the convention to be:

```
wikize_refs.py <basename>
```

and by default it would be assumed the input file `<basename>.src.md` exists and will be read and the default output file would be named `<basename>.md`.

But if the user gives the `--outfile` (`-o`) option, then the user can call with:

```
wikize_refs.py <basename> -o <basename>-wikize.md
```

where it is assumed that the input file is `<basename>.src.md` or the user can be explicit with the input and output with:

```
wikize_refs.py <basename>.src.md -o <basename>.md
```

or with:

```
wikize_refs.py <basename>.md -o <basename>-wikize.md
```

or whatever.

The rule would be if the input name `<inputfile>` from `wikizize_refs.py <inputfile> ...` does not have the `.md` extension, the input file is assumed to be `<inputfile>.src.md`.  But if it does have the `.md` extension then `<inputfile>` is used as the input file name with no name changes.

This would be simpler in the implementation and less surprising the users I think.  And this is kind of how latex works, right?  If you call:

```
$ pdflatex <some_name>
```

it is assumed the input file is called `<some_name>.tex` and the output file that is generated will be called `<some_name>.pdf`.  Nice a simple.  But in our case, the input and output files both are just Markdown and therefore should have the `*.md` extension since that will allow GitHub to format them properly when views on GitHub.

What do you think about that?

Articles/Blog/wikize_refs.py
```diff
@@ -0,0 +1,265 @@
...
+    Parses arguments to wikize-refs.py.
+    """
+    parser = OptionParser(usage())
+
+    parser.add_option("--warn",
+                      default=False,
+                      action="store_true",
+                      help="Warn instead of error during sanity checks.")
+
+    parser.add_option("--no-rdonly",
```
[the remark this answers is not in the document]
markcmiller86: Ok, thats coming in next commit

Articles/Blog/wikize_refs.py
```diff
@@ -0,0 +1,265 @@
...
+    #
+    outlines.append("\n<br>\n\n")
+    for l in comment_lines:
+        outlines.append(l)
+    outlines.append("\n")
+
+    #
+    # write warning comment about this being auto-generated
+    #
+    outlines.append("<!--- WARNING: Auto-generated with wikize-refs.py from %s --->\n"%mdfile)
```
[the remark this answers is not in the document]
markcmiller86: Ok, coming in next commit

Articles/Blog/wikize_refs.py
markcmiller86: Ok, I guess I misunderstood your previous comments on this. I have adjusted the logic now.

markcmiller86:
ping @bartlettroscoe

bartlettroscoe:
@markcmiller86, sorry, I was on travel last week to ICSE19.  I will review this updated script at the same time that I use it to put in the PR for #360 (which just completed SNL R&A).

markcmiller86:
No worries. Just wanted to make sure it was still on your radar. Your plan sounds good.

rinkug:
@markcmiller86 : Mark, I am very curious as to what exactly does this file do? Maybe we can discuss offline, rather than on this page, if you want?

markcmiller86:
@rinkug ... it supports [this refernce styling](https://github.com/betterscientificsoftware/betterscientificsoftware.github.io/blob/master/Articles/Blog/ReferencesInMarkdownHybridApproach.md)

bartlettroscoe:
FYI: Looking at this now ...

bartlettroscoe:
@markcmiller86, looks like the script is a little fragile.  If you don't have the top `- ` text it fails to create the references correctly.

Also, if you run it like:

```
$ ./wikize-refs.py -o WhenIsBetterToWriteAutomatedTests.md WhenIsBetterToWriteAutomatedTests-src.md
```

and then try to run it again you get:

```
$ ./wikize-refs.py -o WhenIsBetterToWriteAutomatedTests.md WhenIsBetterToWriteAutomatedTests-src.md
Traceback (most recent call last):
  File "./wikize-refs.py", line 260, in <module>
    main()
  File "./wikize-refs.py", line 253, in main
    write_output_file(outlines, mdfile, vopts['outfile'], vopts['no_readonly'])
  File "./wikize-refs.py", line 215, in write_output_file
    with open(outfname, 'w') as outf:
IOError: [Errno 13] Permission denied: 'WhenIsBetterToWriteAutomatedTests.md'
```

The script needs to make the readonly file non-readonly, write it, then make it readonly again.

This is screaming for an automated test suite.  Once I post the PR for #360, I will try to write a few quick tests for this script on this branch to show what the issues are.

bartlettroscoe:
@markcmiller86, another suggestion.  It might be better to have the original document `<base>-src.md` to have 1-based references like the generated `<base>md` has.  It is confusing for a reference to change from 4 to 5.

markcmiller86:
> The script needs to make the readonly file non-readonly, write it, then make it readonly again.

I completely disagree with this. There is no way the script can know that the read-only file it was passed the user really for sure wants to write to. You are assuming it is because your workflow is re-running it multiple times with the same target (output). It is the user's responsibility to change the read-only file to read-write if the script is to write to it.

markcmiller86:
> @markcmiller86, looks like the script is a little fragile. If you don't have the top `- ` text it fails to create the references correctly.

I cannot reproduce. What did it wind up doing?

markcmiller86:
> @markcmiller86, another suggestion. It might be better to have 1-based references 

Hmm. I am confused. It does use base-1 indexing for the references. It *does* re-number and potentially re-order the references but that is harmless and since one doesn't edit the result not too important.

bartlettroscoe:
@markcmiller86 said:

> Hmm. I am confused. It does use base-1 indexing for the references.

I guess I was confused as the file:

* https://github.com/betterscientificsoftware/betterscientificsoftware.github.io/blob/master/Articles/Blog/ApolloGuidanceComputerPart1.md

has the first reference starting with `[0]` but the generated file:

* https://github.com/betterscientificsoftware/betterscientificsoftware.github.io/blob/master/Articles/Blog/ApolloGuidanceComputerPart1-wikized.md

has the first reference starting wtih `[1]`.

bartlettroscoe:
@markcmiller86 said:

> There is no way the script can know that the read-only file it was passed the user really for sure wants to write to. 

Okay, I can accept that.  Given that people will be editing the generated file after the fact, we likely need to commit a writable copy.

markcmiller86:
> Okay, I can accept that. Given that people will be editing the generated file after the fact, we likely need to commit a writable copy.

We could add yet another CL switch to force overwrite (which would be off by default) to make the workflow you need a little easier.

markcmiller86:
> @markcmiller86 said:
> 
> > Hmm. I am confused. It does use base-1 indexing for the references.
> 
> I guess I was confused as the file:
> 
> * https://github.com/betterscientificsoftware/betterscientificsoftware.github.io/blob/master/Articles/Blog/ApolloGuidanceComputerPart1.md
> 
> has the first reference starting with `[0]` but the generated file:
> 
> * https://github.com/betterscientificsoftware/betterscientificsoftware.github.io/blob/master/Articles/Blog/ApolloGuidanceComputerPart1-wikized.md
> 
> has the first reference starting wtih `[1]`.

Those were created with an older version of the script

bartlettroscoe:
> We could add yet another CL switch to force overwrite (which would be off by default) to make the workflow you need a little easier.

Given the safely argument, it is likely okay the way it is.  I just run:

```
$ rm -f <base>.md ; ./wikize-refs.py -o <base>.md <base>-src.md
```

bartlettroscoe:
@markcmiller86, I have been using this updated script to produce #377.  (I had to copy the file out of this branch so I could use it on that branch).  It works for my use case which is:

```
$ rm -f WhenIsBetterToWriteAutomatedTests.md 
$ ./wikize-refs.py  \
   -o WhenIsBetterToWriteAutomatedTests.md \
    WhenIsBetterToWriteAutomatedTests-src.md
```

After that, I can provide instructions on how to reproduce the defect I saw.  And we can create a new topic branch with the beginnings of some automated tests (including one that demonstrates the defect I saw).

Okay?

markcmiller86:
Sure. I have run into a bug processing files with xml-style comments in them. I haven't had time to fix it.

markcmiller86:
Ok, I've removed comment block processing from the script entirely. I don't know why I even added it. Maybe to avoid processing references already in a comment block? If so, I can add logic to handle that correctly.

markcmiller86:
Ok, I've adjusted script slightly to monitor whether it is in or not-in a comment block and if it is inside one, it disables processing of references. This permits one to comment out references and the script will still work.

PR #370, opened 2019-05-29
Adding issues with images

markcmiller86:


bernhold:
Thanks @markcmiller86 

In my opinion, there is no need to use PRs for internal stuff like this.  Add what needs to be added.

PR #379, opened 2019-06-10
AGC Software Blog Article

markcmiller86:
This is a DRAFT PR.

Its clearly too long.

I will cut down to something more reasonable later tomorrow and then select final images and re-organize to the images repo.

Would appreciate any feedback.


markcmiller86:
I am working on length reduction. I think I can reduce ~30%. Another commit this PM, then refs, final spell check and image adjustments

jarrah42:
In the "The AGC software stack" table you have "%22" rather than "22%"

markcmiller86:
Ok, thanks I fixed that. Will be in next commit.

Mark


"Greg Watson" wrote:


In the "The AGC software stack" table you have "%22" rather than "22%"

markcmiller86:
I think this is mainly done. I've shortened I think to appropriate length but would appreciate another opinion on that (along with suggestion on what to cut if more shortening is needed). I've spell checked, fixed references and wiki-ized it. 

The only remaining step is fixing pictures. I need to put them to into the images repo and update their paths here. All pictures are taken from NASA documents or creatd myself.

curfman:
Mark -- Very nice, lots of great history!  While this is lengthier than most BSSw content, I do not see an obvious place to cut.  Given that we're aiming to release this week on the site and still need edits from Gail, I will merge so we can pass along to her.  We can concurrently incorporate feedback from others.

markcmiller86:
> While this is lengthier than most BSSw content, I do not see an obvious place to cut.

@curfman .. I found some more stuff to cut down a bit further and committed on master. Lemme know if you object.

PR #382, opened 2019-06-13
Update CC Workflow document as per walkthrough notes

markcmiller86:
I've attempted to update CC Workflow as per [walkthrough notes](https://docs.google.com/document/d/1B_eCJFDy5OqPFrJaAkhp2tefujJNH9F2tOlV4pcU_gQ/edit#)

### CC Issue template
* CC issue template instructions were designed around the notion that the actual CC article would be submitted in the issue body.
  * I reduced this to very basic instructions, requesting a few pieces of specific details
* I removed Pat as automatic assignee as we haven't had a discussion about this

### Workflow document
* After @rinkug walk-through, it was obvious there were many steps in the workflow where it was not clear *who* was responsible for making that step happen.
* I tried to address above by adding **Somehow** as the designator of that task
* Based on @rinku notes and simply re-reading, I wound up with several **Somehows** and that was not satisfying.
* I feel like I am not sure what we are aiming for in this document now? Is it simply to explain to new **EB Members** how things work? To communicate with **Submitters** how content ideas get accepted or rejected? To document to **Authors** how to create content?
* I feel like it kinda sorta touches on all these issues but not too well either and am wondering what is really essential here.
* It feels like too many different people/roles are involved at too many different points in the workflow. It feels more complicated that issue work on actual software projects I work on.
* More importantly, I feel that if we agreed to two things; a) bi-weekly content meetings in which we simply discuss new issues and b) assigning a *responsible* owner to marshall an issue from acceptance to completion, then while it is important to document how things are marshalled for newbies, it is easier to do that than it has been to get a final draft of this workflow document in place.


jarrah42:
I like the new process for the following reasons:

1. I don't think relying on the EB to deal with requests on an add-hoc basis will work. Scheduling regular work sessions seems the best way to make sure that things actually get done.

2. It separates what needs to happen from how it is going to happen. I don't think we should get hung up on details about things like automation or we'll never get started. 

I think a discussion about whether issues that have not received sufficient interest get closed as wontdo or just left in a pending state is warranted, but otherwise I think this is a good workflow.

Site/CuratedContentEditorialWorkflow_Alt.md
```diff
@@ -0,0 +1,66 @@
...
+   * The issue is considered a candidate for further development and moved to
+     Topic Review and a deadline (milestone) is assigned.
+1. Issues in Topic Review are handled as follows
+   * Any more than 30 days old that have garnered insufficient interest are labeled
+     as `wontdo`. A comment regarding to lack of interest is added the the issue is
+     and closed.
+   * Any issues with sufficient interest are moved to the Development Backlog. An
+     **EB Member** is assigned. An author, if different from **EB Member** is
+     identified (via @ mention) and a deadline is set (via milestone) for completion.
+1. Issues in Development backlog are handled as follows
```
bernhold: I'm wondering if we really need to set deadlines for the development of approved topics?  Of course it is better if they don't linger forever, but is there any real harm?

If we get a standing backlog, this could be a place we encourage people with a few minutes of free time to look for something useful they could contribute.

pagrubel:
Thank you for removing me as an assignee. I believe that happened during a walk through for illustration. I'm wondering about the points where we have wontdo's and also the point about waiting two weeks. 
I like the idea of a bi weekly meeting, and think we could appoint someone to be looking at the pull requests during each two weeks so something that is ready to go early in the period gets the proper attention, i.e. a pull request gets merged in a timely fashion.

markcmiller86:
I've updated the *alternative* CC workflow as per today's discussion. Its easies to view the [current version](https://github.com/betterscientificsoftware/betterscientificsoftware.github.io/blob/markcmiller86-cc-workflow-updates/Site/CuratedContentEditorialWorkflow_Alt.md) of that document but if you wanna leave comments or request revision, please do it here in the PR.

Here is summary of chnages
* Added introductory remarks about the document essentially representing state transition squence of CC issues
* Replaced use of `wontdo` label for content that did not ganer enough interest with a `no-interest` label (which has yet to be created)
* Changed name of *Development Backlog* column here in the document with *In Progress* (thought we have to change the project board for this too)
* Changed use of `insufficient development` label with `no-development`
* Expanded the set of steps *after* PR is approved by reviewer but *before* it is merged to include editorial assistance review, possible author review of ea changes, etc.
* Spell checked and corrected typos

PR #387, opened 2019-07-09
Blog agc part3

markcmiller86:
@curfman here is pretty final draft.


markcmiller86:
@curfman I've submitted PR for final work.

PR #397, opened 2019-07-22
Blog Post: Hello Heat Equation

gonsie:
Authors: Elsa G. and Heather
EB member: Mark M.
**Important: Needs review**

This blog post discusses the Heat Equation as **the** "Hello World!" example for the scientific software community.


gonsie:
Both @heatherms27 and I worked on this article.

markcmiller86:
This looks great!!

I like the idea of a blog post, as you have done here, and a working repository of code where, as you appear to have started [here](https://github.com/betterscientificsoftware/hello-heat-equation). I dunno if bssw org will be the right long-term place for it. It might be. If not, we can always decide where it should go later. A working respository of heat eqn examples can also be used as a substrate for various demonstrations of CI in the ECP-CI work.

I am excited about this. Thanks so much for making this contribution.

One quick issue...`Bar.png` isn't available. Maybe a forgotten `git add Bar.png`? Also, I think blog images all get committed [here](https://github.com/betterscientificsoftware/images) with a filename that helps better to associate the image with the particular blog post it goes with. Also, I think some minor adjustments to the mardown would help make it more vanilla markdown instead of so much embedded html.

When I have a free moment, I will read in more detail and provide some more useful feedback.

gonsie:
Thanks Mark. I added the image to the images repo and update the URL. Most of the html is for the math symbols and subscripts / superscripts, but we're happy to take any changes you have.

pagrubel:
Very nice article. In addition to Mark's idea of using the examples in CI for ECP, I would think that those working on programming models might want to use the heat equation for demonstration and/or for instructing those who are beginning to use their programming model.

Articles/Blog/HelloHeatEquation.md
```diff
@@ -0,0 +1,226 @@
...
+* An initial temperature distribution is given
+* Boundary conditions (established tempuratures at each end of the rod) are known
+* Heat only moves horizontally through the bar
+* There are no external heating/cooling sources
+
+<div style="text-align:center"><img width="75%" src="https://github.com/betterscientificsoftware/images/blob/master/Blog_0719_HeatEqnBar.png" /></div>
+
+To model the temperature distribution of the rod, we can use the 1-dimensional Heat Equation given by the formula:
+
+_&part;u / &part;t = &alpha;&nabla; &sup2;u_
```
markcmiller86: @curfman or @bernhold do we know how equations should be handled in blog posts? I'd prefer we use Latex/MathJax as I use [here](https://xsdk-project.github.io/MathPackagesTraining/lessons/hand_coded_heat/) but that makes some assumptions about the back-end.

Articles/Blog/HelloHeatEquation.md
```diff
@@ -0,0 +1,226 @@
+# The Heat Equation is the "Hello World" of Scientific Computing
+
+#### Contributed by [Heather M. Switzer](https://github.com/heatherms27 "Heather Switzer GitHub Profile") and [Elsa Gonsiorowski](https://github.com/gonsie "Elsa Gonsiorowski GitHub Profile")
+
+"Hello World!" is a prototypical program that is universally used by programmers to illustrate the basic syntax of a programming language. At a glance, it can demonstrate a programming language's syntactic structure and can indicate how easy or difficult implementing a basic program may be (see several [Hello World implementations at WikiBooks](https://en.wikibooks.org/wiki/Computer_Programming/Hello_world)). For the scientific software community, a programming language is only the most basic building block of a program. Scientific software must use various programming models or math libraries to achieve high performance for any particular problem. These communities need a version "Hello World!" to compare programming options. The Heat Equation, a well known partial differential equation (PDE), is the perfect formula for this task.
```
markcmiller86: This intro is a little long. I would like to pare it down a bit. Do you mind if I took a whack at it?

Articles/Blog/HelloHeatEquation.md
```diff
@@ -0,0 +1,226 @@
...
+
+## What is the Heat Equation?
+The Heat Equation, also known as the Diffusion Equation, was developed by Joseph Fourier in 1822. Primarily used in the fields of Mathematics and Physics, the equation shows how heat (or something similar) changes over time in a solid medium. While there are many adaptations for different dimensions and scenerios, new programmers are often first introduced to the 1-dimensional version of the Heat Equation as a tutorial problem. Specifically, the Forward-Time, Central-Space (FTCS) version of the equation is used.
+
```
markcmiller86: `FTCS` really only refers to the discretization phase of the algorithm implementation, not the theoretical exposition of the science problem itself.

Articles/Blog/HelloHeatEquation.md
```diff
@@ -0,0 +1,226 @@
...
+The following description is adapted from [a Trinity University lecture](http://ramanujan.math.trinity.edu/rdaileda/teach/s14/m3357/lectures/lecture_2_25_slides.pdf "One-Dimensional Heat Equation").
+
+Suppose you have a thin metal rod along the x-axis starting at position 0 and ending at position _L_. The goal is to model the temperature distribution throughout the rod at a given time, with the following (ideal) assumptions:
+
+* An initial temperature distribution is given
+* Boundary conditions (established tempuratures at each end of the rod) are known
```
markcmiller86: I am curious. I know the rod *feels* more one dimensional but I wonder if that is better than a *wall* which we consider infinite in extent and so focus only on the dimension *through* the wall.

Articles/Blog/HelloHeatEquation.md
```diff
@@ -0,0 +1,226 @@
...
+
+_&part;u / &part;t = &alpha;&nabla; &sup2;u_
+
+Where:
+* _u_ = Temperature at a position, _x_
+* _t_ = time
+* _&alpha;_  = thermal diffusity of the material
+* _&nabla;_ = The LaPlace operator
+
+### FTCS: Assumptions and Discretization
```
markcmiller86: I think its best to go from science problem (the rod), to abstract math formulation (the equation) to discretization (the "mesh") and finally to algorithm (the FTCS update procedure). It feels like we're kinda muddling the last two steps a bit and I think it makes some sense to decouple them more here.

Articles/Blog/HelloHeatEquation.md
```diff
@@ -0,0 +1,226 @@
...
+
+By taking a continuous PDE and turning it into a discrete formula, parallelism can be accomplished. For example, the equation for finding the tempurature at a given point in time in the 1-dimensional Heat Equation is:
```
markcmiller86: IMHO, parallelism isn't so much a result of discretization as it is dividing the work of the algorithm into independently computable tasks.

Articles/Blog/HelloHeatEquation.md
```diff
@@ -0,0 +1,226 @@
...
+
+_u_<sup>_k_+1</sup><sub>_i_</sub> = _ru_<sup>_k_</sup><sub>_i_+1</sub> + (1 - 2 _r_)_u_<sup>_k_</sup><sub>_i_</sub> + _ru_<sup>_k_</sup><sub>_i_-1</sub>
+
+with _k_ being the time, and _i_ being a point in our axis. We can see that the temperature of _u_<sup>_k+1_</sup><sub>_i_</sub> is only dependent on temperatures from spots _i_, _i-1_, and _i+i_ from the previous iteration. By having two arrays to store of old and new temperatures, each location for time _k_ is independent and can be distributed among different processors and cores.
+
+As a consequence of PDEs being so expansive, several implementations are available for multiple programming languages and numerical libraries. Users who are new to coding are often first introduced to the Heat Equation as a means of teaching them different programming languages or tools such as MPI and OpenMP. By reviewing these implementations a solid foundation of specfic language's characteristics can be seen.
+
+## Implementations
```
markcmiller86: I *love* all the implementations here. However, for the blog post. I would just list them with links to the actual code.

Articles/Blog/HelloHeatEquation.md
```diff
@@ -0,0 +1,226 @@
...
+  free ( f );
+
+  return h_new;
+}
+```
+
+While this code is almost identical the hand-coded example shown above, this library does include some extra features that allows users to implement some additional functionality including writing results to different kinds of R8 files, adding a timestamp, and solving for the Courant-Friedrichs-Loewy coefficient.
+
+### More Implementations!
+Using the Heat Equation to demonstrate the power of a programming model or mathematical library is not new. In fact, it is common tutorial exercise for new scientific programmers. Since there are too many interesting examples to include in this article, we've started [a repository showcase](https://github.com/betterscientificsoftware/hello-heat-equation) to collect the many examples.
```
markcmiller86: This is great. Maybe some instructions to people on how to submit a new implementation. We should think about this a bit too. We might want all implementations to share some common base code and behavior.

Articles/Blog/HelloHeatEquation.md
gonsie: @heatherms27 will take a stab at this (I wrote it).

Articles/Blog/HelloHeatEquation.md
gonsie: Good catch. @heatherms27 will review the text to fix-up the wording around the algorithm.

Articles/Blog/HelloHeatEquation.md
gonsie: I really struggled with the wall image that was in your original heat-equation example. The rod makes more sense to me. I can visualize the heat change along the rod... A wall in my mind is a 2D object.

Articles/Blog/HelloHeatEquation.md
gonsie: Will fix up (see earlier comment).

Articles/Blog/HelloHeatEquation.md
gonsie: I guess that it's possible to get parallelization without discretization... Would changing the wording to the following work?

"By taking a continuous PDE and turning it into a discrete formula, it can be more easily parallelized."

Articles/Blog/HelloHeatEquation.md
gonsie: We chose to leave only a few examples in. Do you think removing all the code makes sense? Or, just reducing the examples even more? I feel like it's difficult to talk about the differences in the examples without showing them directly.

Articles/Blog/HelloHeatEquation.md
gonsie: Having a contributing guide is a great idea! I want to stay away from having a common base since this code is not runnable (and making it so would be way too complex). We should point to your CI-examples for runnable code.

Articles/Blog/HelloHeatEquation.md
markcmiller86: No. Discretization and parallelization are wholly orthogonal concepts though it can be argued I guess that a given discretization may enhance or inhibit the work involved in parallelizing it.

I am not sure what you are after here wrt parallelization but the text seemed to imply that *discretization* is what makes parallelization possible. Yes, I guess in that without discretization we cannot parallelize. But the ability to parallelize is rarely (if ever IMHO) the rationale for a given choice in discretization. In fact, I am not sure I am aware of any discretization techniques that we cannot also (easily) parallelize.

Articles/Blog/HelloHeatEquation.md
markcmiller86: Remove all the code except one; mabe the C. List the implementation names and make the names (perma-)links to them in the repo.

Articles/Blog/HelloHeatEquation.md
markcmiller86: I get that, I guess. I just wonder if your experience is rather unique or rather common. I also wonder if there some key explanation missing that would help avoid that struggle for new readers. The reason the wall IMHO is a better model is that it affords some simple extentsions of the problem for more interesting scenarios and helps to define a sciencey problem that is more true-to-life and real-world experience than a simple rod.

Heatherms27: I agree with both statements. In my mind for a simple 1D problem, the rod is easier to visualize with heat just crossing through one axis. The wall example does seem more practical and easier to extend, but it also reminds me of the heated plate example, which seems to be a common when describing the 2D version as heat travels through a mesh structure along the x- and y-axes. (Seen here: http://ramanujan.math.trinity.edu/rdaileda/teach/s12/m3357/lectures/lecture_3_6_short.pdf). My reasoning was that because this article was more focused on the software side of the coin, the rod would be easier to understand, while if it was a more math-centered blog post the wall may be better. That's just my 2 cents though.

Articles/Blog/HelloHeatEquation.md
gonsie: In this *particular* example, parallelization divides the space dimension, whereas discretization divides the time dimension.

Am I expressing the distinction correctly?

markcmiller86: No, not really. In the continuous mathematical equation, both time and space dimensions are *continuous*. To evaluate the PDE with a digital computer algorithm, we must discretize *all* dimensions meaning time and space. The discretization in space is perhaps a little easier to "see" but the discretization in time is equally important to "see". It manifests as the fact that we *evaluate* the PDEs only at specific moments in time...in this case those expressed by `k*delta-t` where `k` is some integer. If we were using a more sophisticated time-integration algorith, we might *vary* delta-t as we go and take bigger steps in time as the dynamics smooth out. But, we'd still be *discretizing* in time. This is part of the reason I ask the reader if s/he can "see" a *mesh* [here](https://xsdk-project.github.io/MathPackagesTraining/lessons/hand_coded_heat/)? Indeed, there *is* a mesh as there and its important to be aware of that fact even if it is subtle. It serves as a launching pad for getting into more interesting (e.g. non uniform) meshing and discretization. Does that make sense?

Articles/Blog/HelloHeatEquation.md
gonsie: @markcmiller86 we tried shortening the intro but didn't make much progress. Please take a whack at shortening the latest version.

Articles/Blog/HelloHeatEquation.md
gonsie: I think we addressed this comment. We wanted to leave in the fact that the heat equation is often used to teach parallel programming, but I think we've decoupled that from the idea of discretization.

Articles/Blog/HelloHeatEquation.md
gonsie: We are working on a contributing guide for that repo. I'll mark this as resolved here.

markcmiller86:
@gonsie @Heatherms27 ... I will be away for ATPESC until mid-next week and won't be following closely. Lemme know by email when you want another look.

Articles/Blog/HelloHeatEquation.md
markcmiller86: Pardon me...I hadn't followed closely enough to understand what changed in the PR vs. responding to questions raised here in the conversation about the PR.

If you think this is resolved (my main concern was to not have a statement that suggests that discretization alone is sufficient to *accomplish* parallelism), then feel free to resolve.

Articles/Blog/HelloHeatEquation.md
gonsie: I'll mark this as resolved since we are able to use HTML symbols, but I raised the issue in #405.

gonsie:
@markcmiller86 We are ready for a final look-over. If you are happy, I can merge the PR (but is that the correct next step for publication??)

markcmiller86:
@gonsie and @Heatherms27 thanks for your efforts moving this forward.

The right next step is to have the BSSw editorial board review for relevance and alignment as well while it is still in this PR. I suspect there will be some modest changes forthcoming from that.

I will move it onto the project board so next board meeting it will be reviewed (I should have probably done that already).

markcmiller86:
I have taken a deeper look at the various implementations cited here and think they are all useful to an extent but that I need is for them to be implemented in such a way that they all accept the same command-line options and output the same result file(s), that all the code needed to compile and run them, including the makefiles (or cmake), are in the repo, etc. In addition, it would be good to have baseline results from key runs of specific science problems, etc.

So, I think if we want this blog article to announce the existence of a repo for people to start populating additional heat equation examples, we have some work to do on the examples we already have in hand before we can publish.

OTOH, if we want this blog article to only introduce the concept of using the one dimensional heat equation as a  "Hello World!" for CSE community, then I think we can revise this substantially (I have already started that), to do that.

@gonsie @Heatherms27 ...  I am not suggesting you need to do any more work here. I am only outlining the reasons I think suggest we should defer publishing this article until we have the aforementioned repository and code in place. Thoughts?

Heatherms27:
@markcmiller86 Thanks for looking over it again! I see your point in having a repository that has runnable code along with an associated make/cmake file. The reasoning behind it just having snippets of code was that we simply wanted to introduce the concept of using the 1D heat equation as a introductory or tutorial example for the CSE community and show the basic syntactical differences, outline several different libraries/tools that are available, and show how hand-coding a simple 1D heat program differs greatly from using a well-developed library such as MFEM. @gonsie may have a differing opinion though. 

_"OTOH, if we want this blog article to only introduce the concept of using the one dimensional heat equation as a "Hello World!" for CSE community, then I think we can revise this substantially (I have already started that), to do that."_

This was more of my take on the article. I'm curious to hear your opinion of how to revise this.

Thanks again!

markcmiller86:
@Heatherms27, I just commited a PR, #423 to this branch with a major revision. You can view by going to the *Files Changed* tab, then going to the elipsis in upper right corner of the text box displaying the diffs and selecting *View file*.

gonsie:
I agree with @Heatherms27, there is no requirement that the code snippets be runnable or accept a common set of command-line arguments. I believe that the code required to implement those things would be significantly more than the code that demonstrates the particular programming model / methodology. As such, the relative ease of comparing the different implementations would be lost.

rinkug:
Continued discussion will occur in the PR #423

PR #398, opened 2019-07-27
Fix listing of questions/criteria (#360)

### LINKED ISSUE #360: "When is it better not to write automated tests?" Blog article

bartlettroscoe:
After the technical edit, the listing of the 5 criteria become a bit awkward.
By changing these to questions, I think this reads better (in my opinion).

Resolves #360.

### DIFF
--- a/Articles/Blog/WhenNotToWriteAutomatedTests-src.md
+++ b/Articles/Blog/WhenNotToWriteAutomatedTests-src.md
@@ -1,8 +1,4 @@
-# Pre-Reference Version: When NOT to Write Automated Tests?
-
-**Hero Image:**
-
-- <img src='https://github.com/betterscientificsoftware/images/raw/master/testing-hero-image.png'/>[Image Source: https://www.testingspot.net]
+# When NOT to Write Automated Tests?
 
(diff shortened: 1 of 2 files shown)


bartlettroscoe:
@curfman, here are the edited files that I think are ready to publish.   Not sure why but the generated file was getting edited instead of the source file.  The only edit that should be made to the generated file is the "Publish" attribute.  All other changes should be made to the the '-src.md' file and then have the file generated with the script:

```
$ ./wikize_refs.py -o WhenNotToWriteAutomatedTests.md WhenNotToWriteAutomatedTests-src.md
```

Otherwise, we are going to write over other changes we when regenerate the file.  (I did this and then noticed that there were other manual changes that I had to go back and fix.)

PR #425, opened 2019-10-04
Finalizing hello world heat equation blog post

markcmiller86:


markcmiller86:
@curfman, @rinkug, and @bernhold this is ready for final review and publication

jarrah42:
I have two comments, otherwise I like the article. First, the title confused me as I don't understand what it means. I would think a simple "Hello CSE World!" might be clearer. Second, after I read the code repository section, I starting looking for a link to the repository. Then I re-read the section about the temporary code snippets and realized that is what you have at the moment. Is that not going to be the repository? Maybe you could be a bit clearer about what you have now and what your plans are for the future?

markcmiller86:
Was trying to be cute with the title. You'd have to be a Rene Zellweger fan to catch the reference. Its from the movie Jerry Maguire where she says "You had me at hello." 😃. I don't need to be that cute though. On the other hand, so much of our content is a little on the dry side.

Regarding confusion over repository plans and what is available now. I will try a re-wording.

jarrah42:
Sorry to be a spoilsport. :-/ I like Rene, but I haven't seen that movie.

curfman:
Thanks, we'll move ahead with copy edits and image treatment.

Issue #330, opened 2019-03-28
EPIC: Improve BSSw contribution process and stability of BSSw site using strict reviewed PRs


bartlettroscoe:
CC: @markcmiller86, @curfman, @bernhold, @maherou 

## Child Issues

* #702
* #738

## Description

Picking up from the conversation in PR #329, this Issue is to consider and scope out a change to the contribution process for the BSSw site through this GitHub repo to achieve:

* Ensure that the 'master' in the BSSw github repo is always deployable 
* Ensure that all contributions are complete are carefully reviewed before being merged to 'master'
* Make it easy to review and provide specific targeted comments about contributions
* Avoid having to manually hack files generated by the `wikize_refs.py` script.
* Teach people the basic GitHub PR contribution workflow in a non-stressful way

The key to achieving by using the basic GitHub PR workflow where the 'master' branch is locked down and all PRs must have a least one other person sign off on them :-)

To allow for previewing articles, one can use an ["integration test branch"](https://docs.google.com/document/d/1uVQYI2cmNx09fDkHDA136yqDTqayhxqfvjFiuUue7wo/edit#heading=h.2r0g9kvx5b2a) called something like 'preview-content' (or using the standard name 'next').  The way this works is that you merge the topic branch for a PR that you want to preview on the website into the 'preview-content' branch and then the BSSw preview site is taken off of 'preview-content' instead of 'master'.  Then work in continued on the topic branch until all reviews are done, all tweaks are complete, etc.  When the topic branch is 100% ready to publish, then the topic branch is merged to 'master' by merging the PR.

Another thing that would improve this process is for the website to automatically not render pages that end with `-src.md`.  That way, all changes to the content could be made in the `<base-name>-src.md` file and used to generate the `<base-name>.md` file using the script `wikize_refs.py`.  (Or we could find another way to address this perhaps but it needs to be addressed if you look at the mess the process was, for example, #377).

The IDEAS-ECP level project board issue for this is:

* https://github.com/bernhold/ideas-ecp-overview/issues/59

# Tasks

* Create PR to modify `HowToContribute.md` with suggested PR process

* Submit one or more curated content submissions as PR to demonstrate the different processes

bartlettroscoe:
FYI: I just created the PR #332 completely on the GitHub site.   That was created in my GitHub fork proving that other users can do this as well, even if they don't have push access to this main BSSw github repo.

I will document how I did that in a PR for an updated `HowToContribute.md` file.

[...]

PR #424, opened 2019-09-27
Create BalterGithubCommunication.md

markcmiller86:
Draft CC article for #375, @rusu24edward is author

PR #436, opened 2019-10-22
Documentation of proposed preview workflow for content PRs

bartlettroscoe:
This documents the `preview` branch workflow proposed in #330.  See the current document [here](https://github.com/bartlettroscoe/betterscientificsoftware.github.io/blob/330-preview-pr-workflow/Site/PullRequestPreviewBranchWorkflow.md).


bernhold:
Would this be useful to automate the merges? https://github.com/bobvanderlinden/probot-auto-merge

Site/PullRequestPreviewBranchWorkflow.md
```diff
@@ -0,0 +1,132 @@
...
+are done un-reviewed on `master`.
+
+
+## Actors in the workflow
+
+* **Author:** Creates a branch with the first draft of the proposed content on
+  a branch (i.e. `content-A` in their fork of the BSSw repo) and then creates
+  a PR of that branch against BSSw `master`.  (This could be actual content
+  author or the editorial board member that collects the initial draft of the
+  contribution from from Google Form placed there by the original author.)
```
bernhold: What is the author experience if they do not have write access to the repo?  They needs to fork, develop their content locally and create a PR for our repo.  But then how does the iteration on the PR work for them? (I have no experience with this.)

Site/PullRequestPreviewBranchWorkflow.md
```diff
@@ -0,0 +1,132 @@
...
+* All edits and reviews to the proposed content are performed on the
+  `content-X` branch and the associated PR before merging to the `master`
+  branch.
+
+* When the proposed content is 100% ready to publish, the `content-X` branch
+  is "graduated" (i.e. is merged to the `master` branch by merging to
+  associated PR).
+
+* The `preview` branch always contains `master` so every merge to `master`
+  must be followed up by a merge of `master` to `preview`.
```
bernhold: The reality is that there are articles that may have been unpublished for some reason.  Right now, we do this by changing the article metadata to publish=preview or publish=no.  In a world where `master` == published and `preview` == preview instead of relying on the metadata, what do we do with unpublished articles?

Site/PullRequestPreviewBranchWorkflow.md
```diff
@@ -0,0 +1,132 @@
...
+
+The commands to update the `preview` branch after the `content-X` branch and
+PR has been "graduated" (i.e. merged to `master`) are:
+
+```
+[(<branch>)]$ git checkout preview
+[ (preview)]$ git pulll      # from origin/preview
+[ (preview)]$ git merge origin/master
+[ (preview)]$ git push       # to origin/preview
+```
```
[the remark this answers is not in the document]
bartlettroscoe: > Are the `[(<branch>)]$` and `[(preview)]$` meant to denote the working directory in which these commands are issued?

No this is the current branch as displayed by the [git shell prompt script](https://github.com/git/git/blob/master/contrib/completion/git-prompt.sh).   I just left out the working directory (because it is supper long for this github repo).

Site/PullRequestPreviewBranchWorkflow.md
bartlettroscoe: > What is the author experience if they do not have write access to the repo? 

If the actual author does not use the fork and PR model, then they can't edit the PR branch.  All they can do is ask the EB member to do it.  This is what we saw in PR #424 and it is very unfortunate.  If the original author is comfortable with github, we need to strongly recommend they fork, create branch, and create PR so they can push updates to the branch.  Other options  are possible but I don't even want to describe them hear to avoid making people's head spin.

Site/PullRequestPreviewBranchWorkflow.md
bartlettroscoe: > The reality is that there are articles that may have been unpublished for some reason.  Right now, we do this by changing the article metadata to publish=preview or publish=no.  In a world where `master` == published and `preview` == preview instead of relying on the metadata, what do we do with unpublished articles?

Two options:

1) Keep using the metadata to toggle publish or unpublish on 'master'

2) Remove the *.md file (using a new commit and PR).

For the same of simplicity, I would go with option-2.  (Note, file is not really gone because it is git history **forever**.)  In any case, there is always going to need to be a new git commit to either toggle the metadata or remove the *.md file so take your pick.  This is not key to the workflow in my opinion.  Either option is okay.

Site/PullRequestPreviewBranchWorkflow.md
bernhold: I understand that this is really the only way for the author to interact.  I was trying to get more at the details of what the iterative refinement process looks like from the author's perspective at the git command level.

bartlettroscoe:
Here are some GitHub APPs taht might be close to what we need to have a PR branch with the `preview` label automatically merged to the `preview` branch:

* https://probot.github.io/apps/deploy/
* https://github.com/dessant/label-actions

But in general it looks like [GitHub Apps](https://probot.github.io/docs/) is the way to go.

You write these with Node.js (and I have never written Java Script).

markcmiller86:
@bernhold thanks for putting together the [test plan](https://docs.google.com/document/d/12aKS-ZNvvDeOzWSAlANUjcQGE9D1hH4kw1jV1XOainY/edit?usp=sharing) and opening a discussion about it. I think today's discussion was good.

There were a few key observations that were made during the discussion that I think we should capture and so I am adding them here as a comment in the PR...

- We have inward facing Git workflow complexity that EB members/staff will be exposed to and then *outward* facing Git workflow issues that all our candidate author's will be exposed to. For example, does anything about this workflow cause Git-naive authors not to be able to work entirely from GitHub *easily*?
  - @bartlettroscoe explains that as long as authors fork and submit PR from fork, life should be good even via GitHub
   - We know of a case of CC content submitted via Google form and then turned into a PR by @markcmiller86 that that route prevents author from being able to edit/revise the PR and this should be mentioned in the how-to-contribute guidance.
- @bartlettroscoe observes that most Git projects use Git command-line for most work and GitHub primarily for managing multiple branches of development and collaborating on PRs related to them. BSSw is, however, aiming perhaps the revers of that...to enable most authors to use GitHub (enabling them to avoid having to learn too much of Git), and GitHub may not be designed too well towards that use case.
- Use of the proposed preview workflow may obsolete a number of things handled currently by metadata blocks. For example...
   - `publish=yes` ==> merge to master
   - `publish=preview` ==> merge to preview
   - Unpublish ==> delete from master
   - `publish=no` ==> not yet merged to master or preview
- Maintaining images in separate repo has some advantages
   - site stylers can make all changes they wan there directly on master and not get in the way of other content development work so unlikely to cause conflicts except
      - So, this isn't as true if the author of an article is also developing the images to be used in the article. @markcmiller86 has had several cases where he's been updating/revising images and those same images have undergone changes by site stylers causing conflicts. Git and GitHub do nothing to resolve this anyways because images are binary content. But, it is an important aspect of images to keep in mind.
   - No image is published until it is referenced from an `md` file on the main site
- We should eventually LSF the images repo
- We should adopt practice of using something like [pngquantify](https://pngquant.org) to reduce image sizes. I've used this tool a lot and it works great...improves image loads speed into browsers too (reduces our GitHub bandwidth quota usage too as does LSF'ing the images repo)
- There are two sort of thinking caps here; Git-purists and Git-pragmatists 😄
   - Note: speaking for myself (@markcmiller86) being a Git-purist does in no way imply any sigificant competence with Git 😄 
  - This possibly raises its biggest head where changes directly on master are concerned
  - We agreed a change to fix a type or some rarely occurring, small, isolated, simple change should be ok to be done directly on master.
  - However, we seem to be considering adopting this new workflow but at the same time including within it the routine need to make larger, wide-sweeping, routine changes directly on master. Examples put forth include recent changes like addition of tables or event contributors. **FWIW...I think this is a recipe for disaster**! If we decide we should adopt this workflow, then we should be committed to handling such changes according to it...make a PR, merge it to preview, test it, approve it, etc. and then merge it to master. Certainly, if we routinely make large changes directly on master, any automation of merges from master to preview will likely cause conflicts requiring human intervention thereby underemining any automation to be gained.

bernhold:
Just one small point about LSF: it is not available in GitHub Pages.  Of course we're not using Pages now.  I don't know if that would ever be feasible for BSSw. And I presume that we could move back away from LFS if we really needed to (at worst we could create a new repo without LFS and copy everything over).  But I just wanted to make sure this point was part of the record.

bartlettroscoe:
> Just one small point about LSF: it is not available in GitHub Pages. 

Ouch.  I did not realize that.  But long term, we many not use a GitHub repo for images at all.   It just occurred to me that you could just store images as public-viewable on Google Drive and then just link to them.  That might be the best long-term solution actually.

markcmiller86:
> Just one small point about LSF: it is not available in GitHub Pages

I don't think prevent us from using GitHub pages for BSSw site if we so chose at a later time. It only means we would be *required* to handle images the way we are currently handling them, in a separate repo.

bartlettroscoe:
One use case that has not been mentioned before is this one:

* PR is created with proposed new content and is merged to `preview` branch but it is decided not to publish to `master` (for some reason)

In that case, after it was decided not publish the content, one would want to close (but not merge) the PR and stop displaying it on the preview site.  To do so, I could see one of three options:

* **option-a)** Set the `publish=on` meta-data in a new commit on the PR branch (then allow the merge to the `preview` site)
* **option-b)** Remove the new `*.md` files and add commit to undo any other changes in PR branch (and then allow merge to `preview`)
* **option-c)** Create a new commit in the `preview` branch to remove the content from the preview site.

There are pros and cons here but I think *option-a* or *option-c* would be best.  That way, you could still see the proposed file in the closed (but not merged) PR.  The advantage of *option-c* is that we could eliminate the `publish` metadata.

markcmiller86:
> The advantage of _option-c_ is that we could eliminate the `publish` metadata.

I was about to write that I vote for option-c and this sealed the deal.

bernhold:
We need to have a way to keep unpublished files around.  Perhaps not all, but some.  

We do this with markdown templates for different article types. These need to be easily accessible/findable by would-be contributors, not squirreled away in some obscure branch..  Maybe an alternative for these is to put them in a separate directory that's excluded from publication.  Though I think it is more natural for users to find the templates in the same directories where there new content will go.

I'm reluctant to rely on git histories to dredge up other unpublished content.  Maybe that's my inexperience with git speaking. But I need to give more thought to other unpublication situations we've encountered.

bartlettroscoe:
> We need to have a way to keep unpublished files around. 

One disadvantage of keeping cruft around is that new authors might accidentally use one of these `*.md` files as an example of how to create a contribution and get sent in the wrong direction.  Think of it as commented out code.

> Maybe an alternative for these is to put them in a separate directory that's excluded from publication. 

I think a `Templates`/` subdir would be a reasonable place to put templates and actually make them easier to find.

> I'm reluctant to rely on git histories to dredge up other unpublished content

Yeah, but the (closed or merged) PRs on GitHub will stay around forever so you can always find stuff there.

markcmiller86:
> We need to have a way to keep unpublished files around.

If the kind of stuff you are talking about here is *adminstrative* stuff such as workflow documentation, etc., that stuff doesn't necessarily need to stay *with* the stuff that is being published. They could be in a separate repo. Call it *bssw-admin*. I am not saying we *should* do that. Only that it is one way of satisfying the requirement *without* resorting to metadata blocks.

The potential for adopting new processes that do not send us down the path of continuing to *require* metadata blocks is important I think. I think we should be working in a direction that frees us of this encumberance. As an aside, I've never understood why this is not presently handled in the Jekyll-conventional way; as YAML *front-matter* instead of an XML-style comment. If it were handled in the *Jekyll* way, I think I would have less issue with it because it means any one of us accustomed to *Jekyll* and *Liquid Templating Language* could make any revisions we deem necessary in the future and thereby also frees us of an external resource dependency.

bernhold:
> As an aside, I've never understood why this is not presently handled in the Jekyll-conventional way; as YAML front-matter instead of an XML-style comment. If it were handled in the Jekyll way, I think I would have less issue with it because it means any one of us accustomed to Jekyll and Liquid Templating Language could make any revisions we deem necessary in the future and thereby also frees us of an external resource dependency.

I agree that it would be desirable to transition to a YAML front-matter approach, as you suggest. One of the other reasons that I like it is because the approach makes the metadata *visible* in any normal markdown viewer.  One of the reasons I think the metadata is so mysterious to most authors and EB members is that it is so well *hidden* today. (There are other reasons too, IMHO.)

I can't speak to the history, but I raised this with Sandbox/Parallactic quite a while ago, and they said that it could be done if we wanted it.  We had other priorities and didn't get into the level of effort or development time or anything like that.

Also, I think the idea of YAML front-matter and Liquid, etc. are separable.  Changing the metadata processing should be much more confined from the perspective of the backend.  I like Jekyll and Liquid and all too, and I feel like a great deal of the site design and functionality could be recast using those tools.  But that would be a much more significant change.  And I think at the moment, we're better off looking for more incremental changes that take us in the right directions.

bernhold:
Regarding the unpublish discussions, we need @curfman and @rinkug to think about what makes sense and what they're comfortable with.

The proposals on the table that would deal with unpublished content that would also allow us to move away from the need for a `publish` metadata element are (@bartlettroscoe @markcmiller86 please correct if necessary):
* Markdown templates for the various content types need to be moved out of the content directories into another location (e.g. a `Templates/` directory) that would be excluded from publication.
  - I think we have other directories that are excluded from publication, so I don't think that's a big deal
  - But a real disadvantage (to me) is that the templates are moved away from the place where the completed contribution needs to be added.  I think that could be confusing to contributors.
* Other content that needs to be unpublished is deleted (ideally via another PR).  It would be recoverable through the git history.
  - For my part, I need a better understanding of how this works to be comfortable with it.  E.g., searching for unpublished content, etc.

bartlettroscoe:
This PR is likely way out of date.  Instead, we can focus on #330 and #517 

Closing

PR #468, opened 2020-02-19
adding first draft of rse podcast blog post

vsoch:
hey @bernhold ! Here is a first shot at an article about RSE Stories - please take a critical look and let me know if it's too simple (it just says how/why I created it, and then how I think of an RSE). I can definitely take a different route, or address a different set of questions. The image for it I added as a PR here https://github.com/betterscientificsoftware/images/pull/5.

Thanks for inviting me to write here! It definitely will be very helpful to reach folks that are RSEs to tell their story, and to create awareness for RSEs for those that aren't.

Signed-off-by: vsoch <vsochat@stanford.edu>

### DIFF
--- a/Articles/Blog/RseStoriesPodcast.md
+++ b/Articles/Blog/RseStoriesPodcast.md
@@ -6,7 +6,7 @@
  
 #### Contributed by [@vsoch](https://github.com/vsoch "@vsoch on GitHub")
 
-#### Publication date: March xx, 2020
+#### Publication date: April xx, 2020
 
 ### We look to experts for knowledge, but do they know the whole picture?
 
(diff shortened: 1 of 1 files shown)


vsoch:
Some general thoughts (and I'll add more as they come up):

 - the paragraph about how it's done (e.g., audacity, github pages) is sort of boring, but I suspect people will want to know these things. Should this be added to some appendix / smaller section instead?
 - probably should mention the target audience somewhere
 - possibly something about it being timeless. For example, I suspect many students that are interested in research software might listen to understand career paths (although uncharted) they might take. Whether they listen now, or in 5 years, a lot of the stories / advice will still hold true.

bernhold:
Hi @vsoch , sorry for the delay in providing feedback.  I think you're off to a good start.  A couple of suggestions. 

Consider swapping the first two paragraphs (or maybe just dropping the first?).  I think second para is a better hook for the reader.

For the last para, consider something along the lines of "What I've learned about RSEs from my podcast"

vsoch:
All set.

bernhold:
Hi @vsoch , sorry about the lengthy delays here.  We had a couple of articles that we needed to get out by particular times.

I like the latest version of the article.  The one thing I would request is to add a bit about your feed and how you're advertising it.  I think that's an important bit to add.  I don't remember exactly where I first heard about RSE Stories, but I do remember that it didn't seem that easy to find out about new episodes, etc.  Now, of course, it is much better.

Once that's in, I want to get one other quick review and then hand it to our technical editor.  Our design team is working on a hero image.  I've asked them to key off of your coffee cup motif and/or a stock image that shows an interview for a podcast (or radio broadcast).  I'll let you know when they have something for us.

vsoch:
okay all set! I added a section, and can answer your question here as well. I don't advertise it as you'd think of how a marketing department does advertising, but I do publish the feed on itunes (and this makes it available to really any podcast service you use) and then I announce episodes on Twitter and in various RSE related slacks.

PR #490, opened 2020-03-17
Add extra blank line before last comment

markcmiller86:
Markdown uses blank lines to differentiate between different contexts.

The old code was putting the `WARNING: DO NOT EDIT! Auto-generated...` message just after the end of the table of references and this was breaking the table for *normal* GitHub flavored markdown renderers.

This update just adds the needed extra blank line


markcmiller86:
Oops. I caught a couple of other files I hadn't intended here. Closing and will re-open a new one

Issue #489, opened 2020-03-17
Working Remotely in an Agile World


bartlettroscoe:
Author: Ross (Written by Rinku)

This is a 48 minute video on "Working Remotely in an Agile World":

https://www.infoq.com/presentations/practices-working-remote/?utm_source=email&utm_medium=culture-methods&utm_campaign=newsletter&utm_content=11262019

This was highlighted in one of the InfoQ summary sent out a few months ago.

bartlettroscoe:
I am watching this now and will post my notes here.

bartlettroscoe:
CC: @rinkug, @markcmiller86, @curfman, @maherou, @elaineraybourn 

My notes from this 48 min presentation are below.

In summary, I think there is some good advice in this presentation and the speaker addresses some issues that I have not seen addressed before like "water cooler" moments and she covers many of the other challenges in communication in a distributed team.  So I think I would recommend publishing a curated content article for this.

One of the big drawbacks was that she asks lots of questions from the audience and requests feedback during the talk, but you can’t really hear what they are saying because they are not given a microphone.

----

**My notes on the presentation:**

* Remote teams:
  * Enhances access to talent for people that will not move to your location.
  * Teams work well by adopting effective communication techniques.
    * Collocated people who are talking need to pause frequently and see if remote people have something to add to the conversation.
    * Remote people should be able to interrupt to get a point across and collocated people should not be offended by this because the lag makes it looks rude.
  * Leadership is not reliant on location.
    * Don’t micro-manage remote people
    * Trust: By default, assume that remote people are working and want to contribute
    * Be flexible to meet at odd times to accommodate different timezones. 
* Invest in Tech:
  * Get good audio/visual-conferencing systems (a different screen for every office and team-room).
  * Do regular one-on-one meetings with remote people.
* Have team days:
  * Get everyone together in the same location on a regular basis
  * Do these whenever there is a change in the team (new team members)
* Error on the side of over-communicating:
  * Reconfirm that was said
  * Rule of 7 (repeat it 7 times)
* Create steams of work (channels):
  * Avoid getting spammed by too many messages (i.e. Slack)
* Have each developer enable an always-on camera with an active speaker so that you create a virtual team room.  (Be mindful you are always on camera and be careful to turn it off when you are away from work.)
* Spend more time preparing for interactions with other team members than you would if you are co-located.
*  Find ways to create “water cooler” moments.
[...]

PR #502, opened 2020-04-02
Remote working in agile world  RemoteWorkingAgileWorld.md

rinkug:
Author: @bartlettroscoe 
EB member: Lois

Based on issue #489


rinkug:
> It occurred to me that we could turn each of these sentence fragments into full sentences and add a period at the end of each by just changing one word in each bullet like "Taking care of" to "Take care of" and then end with a bullet.) Should we make that change? I can do that if you agree.

I think thats a good idea. Please go ahead and do it.

bartlettroscoe:
@rinkug, I switched to full sentences.

PR #451, opened 2019-12-16
Create LackOfEvidenceInSoftwareEngineering.md

jarrah42:
Author: Greg

PR for issue #420 (Lack of Evidence In Software Engineering)

Issue #420, opened 2019-09-19
Lack of Evidence In Software Engineering


jarrah42:
Author: Greg

IEEE Computer has a great article titled "Methodological Irregularities in Programming-Language Research" that everyone should read. In particular, I quote "... [a review] of the evidence standards of research presented at the International Conference on Functional Programming from its founding in 1996 to 2014 and found that not one study followed methodological guidelines... Nevertheless, this research community has vehemently and successfully argued for the inclusion of functional features in popular programming languages such as Java, C++, and Snap!"

https://ieeexplore.ieee.org/document/7999115

markcmiller86:
FWIW...I find this far to the abstract end of the nuts-n-bolts software producitivity spectrum. To be clear, it *is* interesting. But, I want a mix of nuts-n-bolts and abstract content of about 70/30.

markcmiller86:
Also, do we *need* to provide rationale for our votes? I mean, I could maybe understand if my vote was somehow special. But, with a sufficient number of votes, the *average* should be sufficient measure of a proposed idea's merit for site without having to engage in discussion about it.

bernhold:
My thinking is that, in the long run, Topic Review should be a quick thing that shouldn't require a lot of discussion or explanation.

But right now, we're trying to figure out what we want BSSw to be when it grows up, and at this stage, I think the discussions are probably necessary to developing a consensus on the kind of content we want to have.

So right now, I feel like it is fair to ask for rationales and otherwise discuss.

jarrah42:
> Also, do we _need_ to provide rationale for our votes? I mean, I could maybe understand if my vote was somehow special. But, with a sufficient number of votes, the _average_ should be sufficient measure of a proposed idea's merit for site without having to engage in discussion about it.

The average would work if everyone voted, but I don't think we're at that point yet.

jarrah42:
> My thinking is that, in the long run, Topic Review should be a quick thing that shouldn't require a lot of discussion or explanation.

I was in agreement until it was my suggestion that got down voted. :wink:

rinkug:
@prwolfe has interest in reading this and following up

rinkug:
PR #451 has been opened for this issue

PR #516, opened 2020-05-14
Correct formatting

prwolfe:
PR for issue #420
Prior PR #451 

I reviewed this in light of the new style guide @rinkug put together and this should match better.

Notes: 
* @jarrah42 should look at/review the two sentence intro in the top deck
* It was not clear that the Decks info on Category/Topic would be populated from the meta-data rather than needing to be directly present.
* Publication date is not as clear as it could be. Is this date of publication of the content being curated or date of publication to BSSw. 

Overall very nice to have something to reference.

Paul

### DIFF
--- a/CuratedContent/LackOfEvidenceInSoftwareEngineering.md
+++ b/CuratedContent/LackOfEvidenceInSoftwareEngineering.md
@@ -1,5 +1,21 @@
 # Lack of Evidence In Software Engineering
 
+#### Contributed by [Greg Watson](https://github.com/jarrah42)
+
+Computer professionals spend significant effort creating new languages for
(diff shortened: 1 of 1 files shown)


pagrubel:
The file name does not match the article name and the article name should be the link not IEEE Computer.

prwolfe:
Missed that title - thanks. Not sure what you mean about the file name vs. article name. I see 

Title: Lack of Evidence In Software Engineering
File Name: LackOfEvidenceInSoftwareEngineering.md

Which is only different in capitalization and removing spaces.  I did not see this in the style guide either - did I miss it?

rinkug:
I added the proposed changes to the file but not through this PR. So, I am closing this.

PR #533, opened 2020-06-08
Fix broken readme links

zdroid:
Updates two links to point to the right documents.


markcmiller86:
Thanks so much @zdroid

PR #537, opened 2020-06-15
Updated to Blog: Lazy Approach to Developing Scientific Research Software

rinkug:
Pre-publish review.

### DIFF
--- a/Articles/Blog/2020-05-lazy-approach.md
+++ b/Articles/Blog/2020-05-lazy-approach.md
@@ -10,7 +10,7 @@
 
 Lots of fancy features may look attractive, but we rarely use all of them.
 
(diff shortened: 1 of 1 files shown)


rinkug:
This review is still in progress.

rinkug:
* "Lots of fancy features may look attractive, but we rarely use all of them" : needs to be revised or replaced with a short 1-2 sentence blurb that segways into the rest of the article
* "Include a license (ideally permissive) such that not only the original author is able to reuse the code at a later stage" : Revise to "Include a license (ideally permissive) such that people other than the original author are able to reuse the code at a later stage."

rinkug:
This PR should be ignore and the article has  already been published

Issue #540, opened 2020-06-18
Possible Blog Article: Experiences porting code to the HIP ecosystem


hartwiganzt:
Witnessing an explosion-like diversification in hardware architectures, hardware portability and the ability to adopt new processor designs has become a central property characterizing software sustainability. In this blog article, we report and discuss the experience of porting CUDA code to AMD's Heterogeneous-Computing Interface for Portability (HIP).

Maybe a little too early, I already formulated the complete blog article [here](https://github.com/hartwiganzt/betterscientificsoftware.github.io/blob/master/Articles/Blog/PortingToHIP.md).

Pull request #541

bernhold:
Thanks Hartwig! The timing is excellent.  We'll work in getting it ready!

bernhold:
Followups to PR #541

PR #541, opened 2020-06-19
Blog Article: Porting to HIP

hartwiganzt:
This is the pull request for issue #540.


pagrubel:
Nice article and very timely. My only concern is linking to the author's pre-print in his github repo, that may evolve or later cause a missing link.

hartwiganzt:
Thanks, @pagrubel. What I can offer is to publish this on arxiv.org. Would that be a good path forward? Or any other suggestions?

rinkug:
Published!

PR #545, opened 2020-06-29
Karbarz-Events

karbarz:
I'm unsure if I did this correctly!
Attempted to create a new event, as per Lois's email.


cferenba:
Hi @karbarz and @rinkug :  thank you for posting this!  It was on my to-do list to post an announcement here, and I was pleasantly surprised to find you had beaten me to it.

PR #553, opened 2020-07-10
Karbarz howtos

karbarz:
all how tos/what is documents transferred to .md format


rinkug:
The contents of this PR could noot be merged directly; however I picked and commited the "how-tos" as a part of another commit.

PR #592, opened 2020-10-15
Thought leader blogroll curated content article

bernhold:
This PR resolves #474.

I decided that Titus Brown's blog (Ivory Idyll) did not have a high enough density of software-related (as opposed to research-related) content to make the grade.

PR #597, opened 2020-10-22
Changes to: Scientific bloggers to follow curated content

rinkug:
Author: @bernhold David Bernholdt
EB membre: @rinkug Rinku Gupta

Opening a  new PR to propose some changes to the #592


bernhold (changes requested):
I understand where you're coming from, but I prefer the original title.

PR #692, opened 2021-01-31
Event for SIAM CSE21

bernhold:

PR #693, opened 2021-02-01
SIAM CSE21 - Identified Deck Text

karbarz:
Regarding PR #692 

Identified deck text in file.
Several profile links are not completed and only link to GitHub, rather than an individual's page.

@rinkug @curfman


bernhold:
The contributors section is known to be incomplete at this stage.  We're trying to decide whether there are too many contributors.

Issue #712, opened 2021-02-16
Remove "categories" metadata tag


rinkug:
The "categories" tag in metadata is a remnant from previous versions of the BSSw website. It is no longer used. The "topic" tag in metadata is used to display the corresponding categories that they belong to. 

@karbarz : Please create one PR for:
- [x] Remove the section on categories from the[ EB docs metadata page](https://betterscientificsoftware.github.io/bssw.io/bssw_content_metadata.html)
- [x] Remove the word "categories" from other locations where it is not needed. [This page](https://betterscientificsoftware.github.io/bssw.io/bssw_content_publishing.html) is one such example.

@karbarz : Please create another single PR for:
- [x] There are content examples files mentioned on the EBdocs website in several places. These example files can be found on the [cc](https://betterscientificsoftware.github.io/bssw.io/bssw_styling_curated.html), [original content](https://betterscientificsoftware.github.io/bssw.io/bssw_styling_originalarticles.html), [events](https://betterscientificsoftware.github.io/bssw.io/bssw_styling_event.html) pages, mostly. We need to remove the categories metadata from each of the files that serve as an example. 
Following is an example of what the metadata looks like
< !---
Publish: yes
Pinned: no
**Categories: skills, development** <-- THIS LINE NEEDS TO BE REMOVED
Topics: revision control, development tools
RSS update: 2021-02-16
--->

karbarz:
PR #713 has been created for the first requested portion of edits, focusing on removing reference to the categories tag on the EB Docs site-specific pages.

An additional PR will be created for removing the categories from the metadata of examples listed on the EB docs site.

karbarz:
PR #714 has been created for the second requested portion of edits, focusing on removing the categories tag in the metadata of examples. 

This issue should now be completed. 
@rinkug @curfman

PR #714, opened 2021-02-18
Removing categories metadata

karbarz:
Removed category metadata from CC example listed on EB Docs site.

### DIFF
--- a/CuratedContent/ClaimsAboutSoftwareEnginScienceReview.md
+++ b/CuratedContent/ClaimsAboutSoftwareEnginScienceReview.md
@@ -24,8 +24,6 @@ The paper provides a compilation of 66 studies related to scientific software de
 <!---
 Publish: yes
 Topics: software engineering, testing, revision control
-Tags: paper
-Level: 2
-Prerequisites: defaults
-Aggregate: none
+Pinned: no
+RSS update: 2019-07-20
 --->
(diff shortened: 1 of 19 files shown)


karbarz:
This is the second PR regarding issue #712. This issue should now be completed. 

The category metadata tag has been removed from all examples listed on the EB docs pages. 
@rinkug @curfman

rinkug:
To discuss: while we are updating these files, we should remove other metadata that is deprecated as well.

bartlettroscoe:
I added this to the BSSW Internal board to make this more visible.

rinkug:
I have removed the deprecated tags and added the 4 tags to the files mentioned as examples

rinkug:
> While we are updating these files, we should remove other metadata that is deprecated as well.
> @kasia, For all the files that serve as examples, lets have only the 4 current metadata in it:
> 
> * Please do the following if any of the 4 metadata is absent in the file
>   ** Publish: should be present, dont change this.
>   ** Pinned: If not present, please add "Pinned: no"
>   ** Topics: should be present, dont change this.
>   ** RSS update: this may not be present in some file. Please add this tag and the date should be the original publication date in the article.
> * Could you please remove "Tags", "Level", "Prerequisites" or anything else from the files?
> 
> Also note that once the files are modified, please check each of the files on the preview website and ensure that the edits did not break anything.

I have removed the deprecated tags and added the 4 tags to the files mentioned as examples.

rinkug (approved):

rinkug (approved):

Issue #624, opened 2020-11-27
Open Research Podcasts


bernhold:
https://www.software.ac.uk/blog/2020-11-27-top-10-podcasts-software-and-open-research

I'm not familiar with any of the podcasts recommended here apart from Vanessa Sochat's RSE Stories, which we had a blog article about.  And there are certainly others in the larger space of interest to bssw.io.  But we might want to consider either adding (selected) podcasts to the existing blogroll article (https://bssw.io/items/scientific-software-bloggers-worth-following) or adding a companion article focusing on podcasts (vs blogs).

bernhold:
Another new podcast: https://codeforthought.buzzsprout.com/

bernhold:
These have been incorporated in PR #716

PR #716, opened 2021-02-18
Update BloggersWorthFollowing.md

bernhold:
Update to incorporate #624


bernhold:
This is an update, and I think it is ready to publish.  Please have a look.

rinkug:
@rinkug : This article has had a title change. Reminder that some prior articles that point to this article need to be edited to point to this new titled article.

PR #718, opened 2021-02-20
Inplace version of wikize refs

markcmiller86:
Addresses/related to #702, #695, #363

PR #725, opened 2021-02-25
Replace sanity with better word

markcmiller86:
Some suggestion to replace the word "sanity" in our documents.


markcmiller86:
I checked repo for "sanity". Found some to replace. This is an inclusive language issue.

There are instances in another PR #718 that are fixed there too.

rinkug (approved):

Issue #702, opened 2021-02-11
EPIC: Working Group: References


bernhold:
Charter: When is it appropriate (useful) to use references in BSSw resources and how should they be styled/presented?

Members: @markcmiller86, @bartlettroscoe, @JeanShuler 

Child Issues: #363, #695, #770

## Tasks

* [x] Discuss and come up with proposal for policies on using references/citations in bssw.io content
* [x] If to allow include of references/citations, determine how references should be handled:
  * [x] Determine if in-text hyperlinks are okay or if systematic references/citations should be preferred?
  * [x] If used, determine the usage of `wikize_refs.py` for managing references/citations:
    * [x] Should, its usage optional and ordained by BSSW?   => If using formal citations, must use references/citations consistent with `wikize_refs.py`.
    * [x] How to handle he source and generated files? => Use in place updates of the references/citations (see PR #718)
* [x] Create a PR that updates the EB docs with the proposed policy and process on the handling of references/citations ...**See PR #750**
* [x] Ask @betterscientificsoftware/bssw-editorial-board members to review PR #750.
* [x] Get back review from @betterscientificsoftware/bssw-editorial-board members for PR #750.
* [x] Move section on citations/references from `bssw_styling_originalarticles.md` to `bssw_styling_common.md` and add sections to `bssw_styling_curated.md` and `bssw_styling_event.md` discouraging formal citations (see #769) ... **See PR #778 ...**
* [x] Run `wikize_refs.py` on all existing `<base>-wikized.md` files, remove matching `<base>.md` files, and rename `<base>-wikized.md` files to `<base>.md` ... **See #779 ...**

bartlettroscoe:
@markcmiller86, @bartlettroscoe, @JeanShuler,

FYI: I added a task list [above](https://github.com/betterscientificsoftware/bssw.io/issues/702#issue-806666108) to guide this effort.

My views on this are given in #363 and #695.  To summarize my opinions on this:
* We should allow references when they make sense.  (For short CC articles, usually not needed since the reference is given at the top.  For articles where there should be more than one reference, don't discourage their usage).
* If using references, the more systematic and less obtrusive references produced by `wikize_refs.py` should be used over hyperlinked text.
* It would be best **not** to commit the generated file produced by `wikize_refs.py` but have it generated as needed (but I realize that may be hard given the current web hosting contractor being used and how the 'preview.bssw.io' site is not getting updated frequently).
[...]

Issue #363, opened 2019-05-24
Some questions and suggestions for markdown references and wikize-refs.py


bartlettroscoe:
@markcmiller86,

I was not aware of the documentation in the file [ReferencesInMarkdownHybridApproach.md](https://github.com/betterscientificsoftware/betterscientificsoftware.github.io/blob/master/Articles/Blog/ReferencesInMarkdownHybridApproach.md) and the script [wikize-refs.py](https://github.com/betterscientificsoftware/betterscientificsoftware.github.io/blob/master/Articles/Blog/wikize-refs.py).    Thanks for providing that.

My first question is if there are any automated test to define and protect the behavior of the script `wikize-refs.py`?  It looks like a non-trivial amount of code and logic.  Where did this code come from?  That amount of code with no automated tests makes me nervous.

My second comment is that this looks really useful.  This is a type of bibtex for markdown pages.  I am wondering if it might be good to name the original source file `<basename>.src.md` and then have the script generate the `<basename>.md` file.   Since this would be useful for any markdown site like a github.io site and not just for BSSw, it would seem like the better display name would be `<basename>.md` instead of `<basename>-wikized.md`.  Or, one could add options to the script to allow the user to decide on the naming conventions and/or explicitly name the input and output files.  It needs to be made clear to people which file should be manually modified by the user.  By making the output file read-only, that helps to make that point.  I see the script also puts in the warning at the top:

```
<!--- WARNING: Auto-generated with wikize-refs.py from ReferencesInMarkdownHybridApproach.md --->
```

so that might be good enough.  (It might be good to puth that warning at the bottom of the file too.)

What do you think?

markcmiller86:
> My first question is if there are any automated test to define and protect the behavior of the script `wikize-refs.py`?

Right. I tested it mainly while developing the article on the apollo guidance computer and so used that as my test.

> Where did this code come from?

I wrote it a few weeks ago. I am not the best python coder. I don't generally think in terms of OO python and it all winds up looking procedural. 


> This is a type of bibtex for markdown pages. 

Right. And, I agree...In fact, I think it has merit as its own little self-contained GitHub project apart from BSSw.

> I am wondering if it might be good to name the original source file `<basename>.src.md` and then have the script generate the `<basename>.md` file.

Good idea. I didn't think too hard about this aspect. I am fine with changing it.

> Or, one could add options to the script to allow the user to decide on the naming conventions and/or explicitly name the input and output files.

[...]

PR #727, opened 2021-03-02
Move wikize_refs.py to utils/, add some tests, change to argparse, edits to doc (#363, #702)

bartlettroscoe:
@markcmiller86 

This PRs make the following changes to the PR #718 (and is pointed to that PR branch) (which addresses #363, #702):  

* Move `wikize_refs.py` to `bssw.io/utils/` dir (see 0c171fb12a6a5ee73836bf9b5b641a062d8ad7b8)
* Add some basic system-level tests for `wikize_refs.py` based on the existing files `bssw.io/Articles/Blogs/ReferencesInMarkdownHybridApproach[-wikized].md`.
* Change from 'optparse' to 'argparse' (please review commit 6862da564b83d8981ce88b8d0c2c34cc6d44b5fe, not the full patch shown in this PR which is not helpful since the file moved).
* Small edits to the document `bssw.io/Articles/Blogs/ReferencesInMarkdownHybridApproach[-wikized].md`

One thing that I could not figure out was the logic about "Handle an isoloated -o / --outfile (which means to use default name of output file)".  I commented that block of code out since the all the basic tests cases that I could figure out all pass.   What use case is this supposed to support and what is the expected behavior?  I need to know that in order to write some tests so that I can modify the argparse implementation to match expected behavior.

One way to do this is to checkout this branch, revert the commit 6862da564b83d8981ce88b8d0c2c34cc6d44b5fe, then add a test cases for when this logic should be executed and pin down the expected behavior.  Then one can revert the revert commit and get the argpase implementation working for these cases as well.


markcmiller86:
Quick question...why so much CMake logic to handle testing of a python tool? I was thinking it would involve just adding another python script that tested it which I anticipated being < 100 lines.

markcmiller86 (approved):

bartlettroscoe:
> Quick question...why so much CMake logic to handle testing of a python tool? I was thinking it would involve just adding another python script that tested it which I anticipated being < 100 lines.

If you are just running a command-line tool and diffing output, a cmake/ctest solution is likely better than using a raw Python test driver.  And with ctest, you can run the tests in parallel, post to CDash if you want (at some point).  If you look at what the system-test driver script does (removes and creates the test subdir, runs the command, print if it fails, diff output files, etc.) then the a python system-level driver would not be much shorter.  The size of those files are just:

```
http://cloc.sourceforge.net v 1.55  T=0.5 s (4.0 files/s, 436.0 lines/s)
---------------------------------------------------------------------------------------------
File                                                      blank        comment           code
---------------------------------------------------------------------------------------------
utils/tests/CMakeLists.txt                                   16              8             61
utils/tests/run_wikize_ref_diff_test.cmake                   24             63             46
---------------------------------------------------------------------------------------------
SUM:                                                         40             71            107
---------------------------------------------------------------------------------------------
```

So there is a lot of blank space and a lot of comments.  The actual test code is around 100 lines all together (for system-level tests).  Adding more system-level test cases involves very little extra CMake code (and Python can't really improve too much on that).

For unit tests, there will be a separate `wikize_refs_unit_tests.py` program that will run as a single ctest test.   I will add the beginnings of that to show how that works.

BTW, I am basically following the process for adding tests to an untested code in:

* http://ideas-productivity.org/wordpress/wp-content/uploads/2016/04/IDEAS-TestingHowtoAddImproveTestinginyourCSESoftwareProject-V0.2.pdf

Here we are on step **3. Add system-level tests** and we are using the "system-level test harness" ctest tool for this purpose.  (CTest is a better choice for a system-level test harness than raw Python IMHO).

The unit test program `wikize_refs_unit_tests.py` gets us into step **4. Add integration and unit tests**.

There is a lot of behavior in this tool that is not currently being tested very well with just these simple system-level tests.  We need to define what types of input is and is not acceptable and what the error messages will be when users screw up (and they will screw up).  But we can add those based on risk (and the risk for various defects is low right now given the nature of this tool and the very low number of current users).

markcmiller86:
Ok, so I guess the short answer to *why CMake?* is that you wanna use CTest, right?

I wonder if everyone here is going to be comfortable with that? I mean, I don't think anyone would object to learning either CMake or CTest with good reason.

I just wonder if introducing these dependencies, as well as any implied demands upon others here to have or develop the expertise to maintain them is a bigger conversation than just getting some level of testing working for this script? I think that is especially true given that there is a simpler alternative...a python test driver script that exits with 0 on success and 1 on any failure and/or issues a useful error message upon failure.

bartlettroscoe:
> Ok, so I guess the short answer to why CMake? is that you wanna use CTest, right?

@markcmiller86, right, cmake is just an easy way to generate the CTestTestfiles.cmake files you need to run ctest.  Also, cmake -P scripts are a nice portable way to run more complex tests (like in this case).

> a python test driver script that exits with 0 on success and 1 on any failure and/or issues a useful error message upon failure

If you would like to replace the `run_wikize_ref_diff_test.cmake` with `run_wikize_ref_diff_test.py` then that is fine with me.  If you want to get rid of cmake/ctest altogether and use the Python unittest module to define and run each test, then that is possible as well (and I have done that too and there are advantages to that as you mention).  In the end, the amount of code will will be pretty similar and will look about the same.  But from experience, you don't just want to check for 0 return codes.  That makes the test pass trivially by putting a return early in the driver or some bad logic that does not check the return code correctly.  Robust tests are better IMHO.

bartlettroscoe:
@markcmiller86, I will post the updated code and system-level tests using cmake/ctest.  If someone want to replace these with a pure Python solution, then that is fine with me.  But in general, if you are running tests, at some point ctest is not bad option somewhere in the devops toolchain (since CDash comes for about free).

markcmiller86:
Should I merge now?

bartlettroscoe:
> Should I merge now?

@markcmiller86, let me update the behavior of the script as per [above](https://github.com/betterscientificsoftware/bssw.io/pull/727#commitcomment-47764879) and push the commits first.  I have the tests updated now I just need to update the code so the tests pass again.

bartlettroscoe:
@markcmiller86, as of commit https://github.com/betterscientificsoftware/bssw.io/pull/727/commits/ab8e033825d611e85c72cdc45425895a822d0e2b that I just pushed, the `wikze_refs.py` script should have the updated behavior we discussed [above](https://github.com/betterscientificsoftware/bssw.io/pull/727#commitcomment-47764879) and should have pretty solid system-level tests for those use cases.

To view the diffs that I made to the `wikize_refs.py` tool, you want to pull this branch and then run:

```
$ git diff HEAD --not 0c171fb1 -- utils/wikize_refs.py
```

Otherwise, you can't effectively see the changes in this PR since the file was moved.  The changes are very minor.

I think this is ready to merge and after that, I think we can merge the parent PR #718.  We can add a unit test harness and the beginning of unit tests later in follow-up PRs.

markcmiller86:
I will be without my machine for the next several hours when I get it back I will do this merge or you are welcome to do it if you prefer. Thanks for all of the work on this!

Sent from my iPhone

On Mar 3, 2021, at 8:48 AM, Roscoe A. Bartlett <notifications@github.com> wrote:

﻿

@markcmiller86<https://urldefense.us/v3/__https://github.com/markcmiller86__;!!G2kpM7uM-TzIFchu!iguvz5EE5dMGR8X9p0Apc-Tee3Ys8kf7jDSStlxX920y_bA9NO1XGC8-adeaJJy1FQ$>, as of commit ab8e033<https://urldefense.us/v3/__https://github.com/betterscientificsoftware/bssw.io/commit/ab8e033825d611e85c72cdc45425895a822d0e2b__;!!G2kpM7uM-TzIFchu!iguvz5EE5dMGR8X9p0Apc-Tee3Ys8kf7jDSStlxX920y_bA9NO1XGC8-add7cO-o6g$> that I just pushed, the wikze_refs.py script should have the updated behavior we discussed above<https://urldefense.us/v3/__https://github.com/betterscientificsoftware/bssw.io/pull/727*commitcomment-47764879__;Iw!!G2kpM7uM-TzIFchu!iguvz5EE5dMGR8X9p0Apc-Tee3Ys8kf7jDSStlxX920y_bA9NO1XGC8-adeBZ_nbYg$> and should have pretty solid system-level tests for those use cases.

To view the diffs that I made to the wikize_refs.py tool, you want to pull this branch and then run:

$ git diff HEAD --not 0c171fb1 -- utils/wikize_refs.py


Otherwise, you can't effectively see the changes in this PR since the file was moved. The changes are very minor.

I think this is ready to merge and after that, I think we can merge the parent PR #718<https://urldefense.us/v3/__https://github.com/betterscientificsoftware/bssw.io/pull/718__;!!G2kpM7uM-TzIFchu!iguvz5EE5dMGR8X9p0Apc-Tee3Ys8kf7jDSStlxX920y_bA9NO1XGC8-addps7V6GQ$>. We can add a unit test harness and the beginning of unit tests later in follow-up PRs.

bartlettroscoe:
Related to my epic SEPW-211

PR #859, opened 2021-05-14
Create 2021-05-RefTest_GPUProgramming.md

sbxchicago:
Test page for references in blog post.

Relates to #702

PR #863, opened 2021-05-16
Example of pure-markdown solution

markcmiller86:
Just adding this as a pure markdown solution. It does a few things "better" than we currently do in `wikize_refs.py`...

* Does ref. list as two column
* Equalizes column widts (with ugly bunch of `&nbsp;` in header
* Does away with using a table column for footnote lables (e.g. numbers) and instead shows those as super-scripts just *ahead* of each ref.

I like this. I think its easy to make `wikize_refs.py` do this. We can of course have `wikize_refs.py` also and the HTML `<div>` items to give back-end CSS styling hints.

### DIFF
--- a/Articles/Blog/2021-05-reftest2.md
+++ b/Articles/Blog/2021-05-reftest2.md
@@ -186,18 +186,9 @@ END ORIGINAL LINK DEFS -->
 
 ### References
 
-<div class="references-wrapper">
-<div class="references">
-<a name="ref1" href="https://www.britannica.com/technology/Moores-law">Moore’s Law</a>
-<a name="ref2" href="https://www.rambus.com/blogs/understanding-dennard-scaling-2/">Dennard Scaling</a>
-<a name="ref3" href="https://www.nature.com/news/the-chips-are-down-for-moore-s-law-1.19338">The Chips are Down for Moore’s Law</a>
-<a name="ref4" href="https://www.omnisci.com/technical-glossary/hardware-acceleration">Hardware Acceleration Definition</a>
-</div>
-
-<div class="references">
-<a name="ref5" href="https://www.top500.org/">Top 500: The List</a>
-<a name="ref6" href="https://www.olcf.ornl.gov/olcf-resources/compute-systems/summit/">Summit: Oak Ridge National Laboratory’s 200 Petaflop Supercomputer</a>
-<a name="ref7" href="https://docs.nersc.gov/systems/cori/">Cori</a>
-<a name="ref8" href="https://docs.nersc.gov/tools/performance/roofline/">Roofline Performance Model</a>
- </div>
- </div>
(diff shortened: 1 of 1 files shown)


bernhold:
@markcmiller86 I can't really evaluate this until I can see it in full preview on the front-end.  What if you made this reftest3 and added it to #859 so that we can compare the options side by side by side?  **Remember to change the title so it doesn't collide with the existing titles!**

markcmiller86:
> What if you made this reftest3 and...

Hmm...I'm a bit reluctant do that because I feel like we're comparing apples (what we can do with GFM on the front-end and what authors would work with) and oranges (what the backend finaly does with the result. Again, I think these should be largely orthogonal in that we can make almost arbitrary choices in either setting and still suit our needs.

I think we need to decide a) how we expect (and will advise) authors to compose content that involves citations and b) how we expect the back-end to render that content and then task contributors with making those things happen. FWIW, I am happy to adjust `wikize_refs.py` to support the authoring end of things.

bernhold:
Mark, I think it matters both how things appear on GH and in other authoring tools *and* how they appear on bssw.io.  Its impossible for them to be orthogonal.  And to me, the goal is something that looks good on bssw.io and is *workable* in the authoring tools.  I think expecting both renderings to be really good is too high a bar.  And I think that favoring how it looks in the authoring tools over the site is just a mistake.

(For example, one admitted disadvantage of Matt's approach is that in the authoring tools, the `<a>` tags for the references all run together in a single paragraph.  I think it would be better if we explicitly put the numbers in front -- at least you could navigate them reasonably in authoring tools, even if the presentation is not ideal.  It might look even better if we were able to express the references an ordered list in Markdown (which I think would just require explicit numbering with a trailing `.`) and the CSS for the front-end could style the list as it likes.)

markcmiller86:
> Mark, I think it matters both how things appear on GH and in other authoring tools _and_ how they appear on bssw.io. Its impossible for them to be orthogonal. And to me, the goal is something that looks good on bssw.io and is _workable_ in the authoring tools. I think expecting both renderings to be really good is too high a bar. And I think that favoring how it looks in the authoring tools over the site is just a mistake.

Maybe you are right. Maybe my expectations of what *should* be possible on the back-end page generation are too high.

FWIW, any other Jekyll-based site I have either used or developed myself enables *wildly* different styling and apperance of the generated/hosted pages from the authored markdown files. The orthogonality of those two parts of the workflow is pretty much why Jekyll themes even exist. Enabling authors to work in all markdown or mostly markdown (a markup language that is highly simplified relative to the HTML that the pages are ultimately rendered into) whilst producing some really nice styling of the generated site is a major advantage of these kinds of sites and content managment.

IMHO, most if not all of what you seem to be concerend with here is solvable by appropriate CSS styling on the back-end and only the back-end. In fact, with a normal Jekyll site and some fancy liquid coding, we could opt to have only single-column references table in the markdown files and nonetheless *still* render that to multi-column, tiny font table on the backend. In proposing to modify `wikize_refs.py` to do that instead, there was an implicit concession by me that for some reasons that are still unclear maybe the bssw.io backend isn't as flexible as I would ordinarily expect.

bernhold:
Mark, this is not a Jekyll-based site per se, but I agree that the principles still apply. What we're discussing here is, essentially, defining the theme for BSSw.io.  It is something that we have some control over, and has evolved over time to meet our needs.  I don't see this as any different.  Matt was trying to demonstrate for us something that is available in the "theme" right now.  Maybe not in the way we'd ideally like it to be.  So we have a chance to try to evolve the "theme" to serve both authors and the published site.  We need to decide first if the kind of presentation that Matt's proposing is desirable, and then how to achieve it.

As I suggested upthread, if we like Matt's styling, I think we might be able to achieve it through the following:
* Authors mark up a References section (`### References`) in the form of an ordered list
* The `### References` tag triggers the front-end+CSS to generate the references styling, which is smaller font and maybe two column.

The required CSS would be somewhat different than what is available in the `references-wrapper` and `references` classes, but not completely dissimilar.  I don't know about the mechanics of flowing it over two columns.  But otherwise

PR #873, opened 2021-05-19
Update FellowshipsApply.md to remove timeline

rinkug:
Timeline was outdated, hence commented out. New 2022 fellowship timeline will be added later.


rinkug:
@hnamLANL : We have updated the fellowship page to remove 2021 timeline

PR #947, opened 2021-07-06
Updating Contributors

karbarz:
Several small fixes will be made to articles, no check list necessary. 

- Error in contributors list. Changed to separate two authors. - SWR 002 Event


karbarz:
Panels and webinars have also been checked and updated as necessary

PR #1072, opened 2021-09-23
Proposing a small change to the workflow

bernhold:
I would like to propose this small change to our workflow.

Currently, once a PR is created, the workflow is to go to the associated issue, document the PR number and close the issue.
However, GitHub will automatically close issues linked to a PR when the PR is merged.  All you have to do is use the right language in the PR to make the link.  I've already changed the PR template so that the language is there (`Resolves #<issue-id>`).  So if people use it properly, the issue(s) will close when the PR is merged.

There are two differences from our current workflow:
1) It eliminates a manual operation that the EB member must do
2) The issue closure is delayed to when the PR is merged instead of when it is created

So there is a bit of a trade-off.  The issue stays around a little longer, but overall we have fewer manual steps.  I think it is worth it.  (And of course there is nothing to stop people from manually closing the issue anyway if they want to.)

I'd like to bring this to the EB for discussion.

### DIFF
--- a/docs/pages/bssw/bssw_curatedworkflow.md
+++ b/docs/pages/bssw/bssw_curatedworkflow.md
@@ -71,9 +71,8 @@ each of the numbered steps in the description below.
      * Manually add to *In Progress* on the *Content Development* project board.
      * Copy details from the issue to the PR: (1) "Assignee" field to indicate EB member, 
        (2) Author name is indicated on first line. 
-     * Close the associated Issue, add comment that continued discussion will occur in the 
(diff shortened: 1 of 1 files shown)


bartlettroscoe:
I agree.  It is fine to keep the Issue open until the PR is merged.

There is a more philosophic argument for keeping the Issue open until the PR gets merged.  That is because the Issue defines the *Requirements* and the PR is the *Implementation* and technically you can have more than one possible candidate Implementation (i.e. PR) that may satisfy the same set of requirements.  But once an implementation is selected and deployed, the issue is complete and can be closed.

PR #1265, opened 2022-03-14
Testing allowing links for the Website item

bernhold:
Instead of just bare URLs.


rinkug:
@bernhold : this seems to work, so can we close?

Issue #1259, opened 2022-03-04
CC: CMake tutorial and reference resources


bartlettroscoe:
I my recent work with a major CMake refactoring effort, I have found the book "Professional CMake":

* https://crascit.com/professional-cmake/

to be incredibly valuable.  I highlighted this book in my SIAM PP22 talk last week:

* https://doi.org/10.6084/m9.figshare.19241715

From slide 5:

* Self-published book:
  * https://crascit.com/professional-cmake/
  * Downloadable PDF
  * 500+ pages
  * Publishes new edition with every major release of CMake (currently 11th edition covering CMake 3.22)
  * Cost: $30 (which includes 10 future editions)
* Covers:
  * Basic CMake features in great detail
  * Modern CMake idioms and best practices
  * Tips and critical behavior (sometimes) missing in official CMake reference documentation.
  * Practices to avoid and why

**Fills a critical gap in CMake tutorial and “why” documentation**

Anyone doing anything with CMake should own this book.

Also, we are pretty thin on [CMake material on bssw.io](https://bssw.io/items/search?search_string=cmake) and ECP surveys have listed CMake help at the top of the list for the last few years.

bartlettroscoe:
@betterscientificsoftware/bssw-editorial-board, can I get some up-votes for this? 

I wish there was a free alternative to this book but there just is not.  Kitware put their book "Mastering CMake" online for free:

* https://cmake.org/cmake/help/book/mastering-cmake/

but it is just not in the same league as "Professional CMake", IMHO.

bernhold:
@bartlettroscoe I advise you to just go ahead.  CC for CMake resources is clearly in scope.

I wonder if it could/should be a single article with a *collection* of resources for CMake (as we have a collection of inclusive language resources)?  Anything that you think is worth recommending.  I know they've done an ECP tutorial or two.  Are the recordings for those available? (Or other people's tutorials), etc.

bartlettroscoe:
> I wonder if it could/should be a single article with a collection of resources for CMake (as we have a collection of inclusive language resources)?

@bernhold, perhaps, but for me, the only two main sources for CMake documentation worth directly consulting and searching are:

* The book "Professional CMake"
* The online CMake reference documentation for your minimum version of CMake (i.e.  https://cmake.org/cmake/help/v3.17/ for TriBITS and Trilinos currently)

[...]

PR #1275, opened 2022-03-20
Fix <img src .../>

bartlettroscoe:
CC: @betterscientificsoftware/bssw-editorial-board 

See diff and commit.

Noticed this while working on #1259


bernhold (approved):
LGTM, good catch!  I would go ahead and merge it.

PR #1339, opened 2022-05-13
Attempt adjust blog article to use refs.

### LINKED ISSUE #1308: Update blog article to use references
https://bssw.io/blog_posts/user-stories-in-scientific-software-development
https://github.com/betterscientificsoftware/bssw.io/blob/master/Articles/Blog/2022-04-user-stories.md

I tried to replace footnote-style references in the original with inline citations, but Matt thinks they're more distracting.  They also cause poor formatting in at least one case (strange line break).  So we should probably change the piece to use wikize-refs citations.

markcmiller86:
@bernhold does this fix #1308 for you?

### DIFF
--- a/Articles/Blog/2022-04-user-stories.md
+++ b/Articles/Blog/2022-04-user-stories.md
@@ -18,7 +18,7 @@ The practice of scientific software development is changing:
     collaboration, and teams becoming more distributed and
     multi-institutional. As a result, there's a greater need for
     **communication and consensus** in scientific software
-    development ([doi:10.2172/1843575](https://doi.org/10.2172/1843575), pp. 176-177).
+    development.<sup>[1]</sup>
 
 -   The forefront of scientific software is increasingly open source,
     open science, and community driven. Most scientific software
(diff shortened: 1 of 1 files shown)


markcmiller86:
Note, I chose to keep two in-line refs because I felt like they were less citations and more FYI's.

markcmiller86:
It seems to look ok to me on preview.

markcmiller86:
@bernhold does this work for you?

markcmiller86:
Hope I am not being a pest...ping @bernhold

markcmiller86:
> markdown within these references isn't rendered properly

I believe we can just re-run `wikize_refs.py` with the `-t` option to do it that way instead (`-t` says to do two-column html).

markcmiller86:
Oh, I guess that is indeed what you did 😉

bernhold:
I did it manually.  But I think it looks okay now (check out the preview if you want: https://preview.bssw.io/blog_posts/user-stories-in-scientific-software-development).  Thanks for this improvement.  I'll merge this shortly, unless I hear back that you have concerns about what you see in preview.

markcmiller86:
Nah. go ahaed if you are happy. Glad it is useful.

PR #1429, opened 2022-09-09
Update links with 2023

hnamLANL:
update links


bartlettroscoe:
Following up from https://github.com/betterscientificsoftware/bssw.io/pull/1423#issuecomment-1247282472 ....

Okay, I see the failed merge to 'preview' at:

* https://github.com/betterscientificsoftware/bssw.io/actions/runs/3025435420/jobs/4867888632

I would have to study the situation here to determine why this filed.  (But note that I have seen times when this GHA merge failed but running git myself worked.  I can't explain that because the GHA does not provide any feedback at all.)

Issue #1377, opened 2022-07-01
Identify and tag existing articles in new topics


bernhold:
We're in the process of adding several new topics to the site (#1378, #777).  Once that's done, we need to review existing articles and tag appropriate ones with the new topics (in the metadata block).

- [ ] Better Planning --> Software Sustainability (note: just because it has the word "sustainability" doesn't necessarily mean it should be included under this topic)
- [x] Better Performance --> Cloud Computing
- [x] Better Performance --> Big Data
- [x] Better Reliability --> Peer Code Review

bernhold:
@markcmiller86 --> Big Data
@bartlettroscoe  --> Software Sustainability
@bernhold --> Cloud computing

Action: PRs with changes to topic metadata

markcmiller86:
I want to mention, in some instances, it might be necessary to add *both* the new **topic** you are assigned to review for and also a new **category**, right? 

I mean, we can't add `big data` as a *topic*, without also ensuring `performance` is one of the categories.

markcmiller86:
I am noticing we have some duplication...

```
% ls Site/Categories/
Collaboration.md	Performance.md		Reliability.md
Development.md		Planning.md		Skills.md
% ls Site/Topics/
Collaboration.md	Performance.md		Reliability.md
Development.md		Planning.md		Skills.md
```

In comparing `Performance.md` in both places, it appears to me the stuff in `Site/Topics` is most up to date. For example, it includes the `big data` topic. It also includes a heading for **Child Topics** as well as relative ordering of those topics.

Is the stuff in `Site/Categories` still used.

bernhold:
The stuff in Site/Categories is not used any more.  Someone was supposed to delete it, but obviously it hasn't happened yet.  I'll do it now.

bernhold:
PR #1468 addresses the cloud computing topic.

markcmiller86:
PR #1463 addresses the big data topic.

rinkug:
@bartlettroscoe : any update on this?

bartlettroscoe:
Thanks for reminder.  I will try to get to this on Friday.

rinkug:
@bartlettroscoe : I took a shot at finding some articles related to software sustainability. Can I create a PR and get this done this week since I have some time?

- https://bssw.io/items/what-is-sustainability
- https://bssw.io/items/software-sustainability-institute-guides
- https://bssw.io/items/an-introduction-to-the-software-sustainability-institute
[...]

PR #1463, opened 2022-10-13
Add big data child topic

markcmiller86:
# Description

EB Member: @bernhold 

Resolves portion of #1377 

This is a minor revision of metadata blocks to appropriate articles.

I took a first whack at this searching for all `.md` files that mention `big data`, `analytics`, or `data science`. Now that I think about it, I should have done from google because my workflow won't catch terms if line break appears between the words.

That said, I wanna make sure I am on the right track regarding following...
* Handling events and articles
* adding `performance` topic if it wasn't previously there before adding `big data`
* Doing very cursory read of article to assess whether it is very relevant to big data.

### DIFF
--- a/Articles/Blog/RSE-HPC-2020Workshop.md
+++ b/Articles/Blog/RSE-HPC-2020Workshop.md
@@ -92,7 +92,7 @@ Daniel S. Katz is Assistant Director for Scientific Software and Applications at
 Publish: yes
 RSS update: 2020-10-26
 Categories: Planning, Collaboration
-Topics: Software Engineering, Projects and Organizations, Conferences and Workshops, big data
+Topics: Software Engineering, Projects and Organizations, Conferences and Workshops
 Tags: bssw-blog-article
 Level: 2
 Prerequisites: default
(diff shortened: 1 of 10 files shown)


bernhold (commented):
Hi @markcmiller86 I would say that just because an article mentions (big) data doesn't necessarily mean that there is any substantive content on the topic provided by that resource.

Of these articles, I would say that the Jupyter and HDF5 webinars are probably the best candidates for the big data topic.  I would not apply it to the others.

bernhold:
Also, for everyone's information the `category` metadata element is not used.  Only the topics are used.  The categories are derived from the topics.

markcmiller86:
@bernhold ok, I've undid the draft changes I've made to the other docs here.

Short of manually going through all files to manually examine them at this point (I was using some grep operations above to narrow scope -- I figured terms like `big data`, `analytics` or `data science` would almost certainly have to appear in any article dealing with that topic in a substantive way so if those terms do not exist in an article, it likely is not a candidate to begin with), do you have any other recommendations for completing this task?

bernhold:
@markcmiller86  that seems reasonable to me.  I based my cloud computing changes on similar searches of the site.  I think it is good enough to go on with.

bernhold (approved):

PR #1617, opened 2023-04-11
Took a whack at some revisions

markcmiller86:
@pagrubel and @rinkug here is a very quick attempt at reducing wordiness a bit. Also, seems like we should list some of the main events if we're going to mention


rinkug:
I think we should skip mentioning events otherwise there may be too many to list...?

Issue #1431, opened 2022-09-10
Periodic CI to find links that have gone bad


markcmiller86:
A lot of what we publish has links. I think links are really important. But, they also go stale over time as other content hosters change how they content we are linking to gets hosted.

We should have something that runs periodically, maybe once a week, and generates a report of bad links.

markcmiller86:
I tried https://www.drlinkcheck.com and got these results. Its a $10 monthly subscription for up to 10,000 links.

### Overview

![Screen Shot 2022-09-10 at 3 31 35 PM](https://user-images.githubusercontent.com/5720676/189504046-c57a10cf-a4c6-4c00-b8ed-a34fce9732a7.png)

### Four bad links found amoung the 1500 checked

![Screen Shot 2022-09-10 at 3 30 57 PM](https://user-images.githubusercontent.com/5720676/189504051-2df90999-fa40-4fed-9a98-b7a6ac640778.png)

bernhold:
Sandbox does this for us periodically.

Though I think it would not be a bad idea to also institute some regular checking within our repo.  I have actions for the bssw-tutorial website that do this on files that change (though this has had some problems lately) and on a scheduled for the whole repository.  See <https://github.com/bssw-tutorial/bssw-tutorial.github.io/blob/main/.github/workflows/check-pr-urls.yml> and <https://github.com/bssw-tutorial/bssw-tutorial.github.io/blob/main/.github/workflows/check-all-urls.yml>.

markcmiller86:
> and https://github.com/bssw-tutorial/bssw-tutorial.github.io/blob/main/.github/workflows/check-all-urls.yml.

Thanks for mentioning. Took a very quick look at [this run](https://github.com/bssw-tutorial/bssw-tutorial.github.io/runs/8218584908?check_suite_focus=true) and it looks like it spews all URLs it checked (some timed out...what does that mean?) and then lists those it think failed. I clicked on some of the failed links and they worked. Something like this is likely sensitive to intermittent issues (in servers, networking, etc.).

We could expand to maintain a list of failed URLs over several successive checks and flag a URL as *bad* only if its gone into a consistent failure state. That would take a bit more work because it would require maintaining a list of the failures across CI runs. But, I think its possible.

bernhold:
[...]

PR #1633, opened 2023-04-18
Create check-published-links-weekly.yml

markcmiller86:
* [ ] Set list of Reviewers (at least one).
* [ ] Add to Project [BSSw Internal].
* [ ] View the modified `*.md` files as rendered in GitHub.
* [ ] If changes are to the GitHub pages site under the `docs/` directory, consider viewing locally with Jekyll.
* [ ] Watch for PR check failures.
* [ ] Make any final changes to the PR based on feedback and review GitHub (and Jekyll) rendered files.
* [ ] Ensure at least one reviewer signs off on the changes.
* [ ] Once reviewer has approved and PR check pass, then merge the PR.

[preview]: https://preview.bssw.io
[bssw.io]: https://bssw.io
[Content Development]: https://github.com/betterscientificsoftware/bssw.io/projects/3
[BSSw Internal]: https://github.com/betterscientificsoftware/bssw.io/projects/2
[meta-data]: https://betterscientificsoftware.github.io/bssw.io/bssw_styling_common.html#metadata-section
[wikize_refs.py]: https://github.com/betterscientificsoftware/bssw.io/blob/master/utils/README.md#wikize_refspy

### DIFF
--- a/.github/workflows/check-published-links-weekly.yml
+++ b/.github/workflows/check-published-links-weekly.yml
@@ -1,0 +1,40 @@
+name: Check published links weekly
+on:
+  # Run every Sunday evening.
+  schedule:
+  # 5:17 AM every Sunday
+    - cron: '17 5 * * 0'
+  # Allow for manually running also.
+  workflow_dispatch:
+
+jobs:
+  urlcheck:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v3
+
+    - name: URL Checker
+
+      uses: urlstechie/urlchecker-action@0.0.34
+      with:
+        # A comma-separated list of file types to cover in the URL checks
+        file_types: .md
+
+        # Choose whether to include file with no URLs in the prints.
+        print_all: false
+
+        # More verbose summary at the end of a run
+        verbose: true
+
+        # How many times to retry a failed request (defaults to 1)
+        retry_count: 3
+
+        # Google Forms is having enormous timeouts
+        timeout: 10
+
+        # Exclude these patterns from the checker
+        exclude_patterns: http://localhost:4000
+
+        # Exclude these files from the checker
+        exclude_files: _config.yml,.github/workflows


bartlettroscoe:
@markcmiller86, please @mention me and let me know when this is ready to review

markcmiller86:
Apologies...I converted to **DRAFT** for time being.

markcmiller86:
Have adjusted it to ignore...

- `docs` -- those links are often not really public...EB should fix but we can do as encountered.
- `Events` -- short lived and links are likely to stale over time

markcmiller86:
@rinkug, @bartlettroscoe and @bernhold this is now ready for review.

The weekly CI check should result in a small list of either bonified broken links or false negatives (e.g. says its broken when its not) to try to check and then either go edit files to fix, add to ignore patterns or just ignore.

rinkug:
This is sounds good. Perhaps we can discuss this during the operation call next week?

markcmiller86:
Summary of kinds of actions taken...

* `fix` the link by finding the associated content at its new URL (most common case)
* `replace` the link to something of similar ilk that served the same purpose because the existing resource is no longer available (one of these cases)
* `remove` the link because the resource was no longer unavailable (one of these cases)
* `ignore` the link by adding to either `exclude_files` or `exclude_patterns` variables controlling the CI logic.

bartlettroscoe:
Can we see an example of what this GHA output looks like?

markcmiller86:
We should enhance to deal with bad links in a pr
pattern exclusions and url exclusions might need to go into sep files.

.github/workflows/check-published-links-weekly.yml
```diff
@@ -0,0 +1,43 @@
+name: Check published links weekly
+on:
+  schedule:
+    # 5:17 AM every Sunday
+    - cron: '17 5 * * 0'
+  # Allow for manually running also.
+  workflow_dispatch:
```
bartlettroscoe, line 7: We would like to add a trigger for PRs to run this on only the files modified in the PR

bartlettroscoe:
Is it possible to put in file-specific ignores?

markcmiller86:
cdc article links change a lot...we should exclude that article.

bernhold:
@markcmiller86 take a look at https://github.com/bssw-tutorial/bssw-tutorial.github.io/tree/main/.github/workflows for examples of URL checking on both a scheduled basis and for PR changed file.  Also there are two files that are used in both actions to exclude certain URLs and URL patterns.

markcmiller86:
remove preview.bssw.io from exclude patterns

bartlettroscoe:
Nice example of operating on only modified files in a PR GHA:

https://github.com/bssw-tutorial/bssw-tutorial.github.io/blob/main/.github/workflows/check-pr-urls.yml

bernhold:
@markcmiller86 you should probably also note that this PR resolves #1431

bernhold:
To connect with our other discussion today, I'll note that I have found that DOI links to some publishers tend to generate false negatives fairly often.  That's why you'll see a fair number of DOI links in the bssw-tutorial exclude list.  On the other hand, because they're DOIs, I am quite confident they will actually continue to exist, so excluding them is a minimal risk.

rinkug:
@bernhold : Whats happening with this PR?

rinkug:
As discussed in todays meeting, @markcmiller86 will get to this month.

rinkug:
@markcmiller86 : does this need a review from me? (or someone else hopefully?)

markcmiller86:
> @markcmiller86 : does this need a review from me? (or someone else hopefully?)

It will when I take out of draft mode. I have several enhancement requests from @bartlettroscoe and @bernhold to add before that. I plan on doing in next week, or two.

markcmiller86:
@betterscientificsoftware/bssw-editorial-board I believe this is near ready now. I can fix the links it is finding in this PR but I thought I'd make a request to have someone look it over.

markcmiller86:
Ok, I still need to re-work this a bit. As it stands now, the list of files to check in the non-PR case would be too long. I know what I need to do to fix this.

markcmiller86:
@betterscientificsoftware/bssw-editorial-board I think this CI is ready.

It does appear to have some false negatives with links to `.pdf` files.  Am checking with developers of urlchecker action.

bernhold:
Is there any chance the PDFs are DOI links?  I often find some DOIs frequently give false negatives with this checker.  I assume it has to do with the redirection and/or a slower site.  I can also imagine that PDFs might take longer to load than most (HTML) links, and time out the checker.

There are parameters you can set for retries and (I think) timeouts and other things that might help.  And if the false negatives are consistent and you're reasonably confident the files really won't go away (like DOIs), you can exclude those files.

markcmiller86:
> Is there any chance the PDFs are DOI links?

Nope. I am in communication with `urlchecker` developers. Its apparently an expired cert issue. I've asked for an enhancement to ignore expired certs. I've also learned the action is not *just* checking links but actually downloading the linked content (like a browser would if you followed the link). This is IMHO bad because it takes **a long** time to validate links.

bernhold:
So it is a site thing rather than a PDF thing, if I understand things correctly?  Consider excluding those URLs. Perhaps with a note (issue) to re-check in a few months.

I found a potentially useful setting for the URL checker today: `verbose: true` will list the files containing the failed links in the summary listing at the end of the job.  No need to search through the repo to back track the source!

markcmiller86:
> So it is a site thing rather than a PDF thing, if I understand things correctly?

Yes.

> Consider excluding those URLs. Perhaps with a note (issue) to re-check in a few months.

Seems too onerous IMHO since bad certs are a common scenario...we'd be doing this all the time. And, the links still "work" in the browser. So, `urlchecker` is just being too pedantic about it.

> 
> I found a potentially useful setting for the URL checker today: `verbose: true` will list the files containing the failed links in the summary listing at the end of the job. No need to search through the repo to back track the source!

Yep. I use that in this CI...very useful.

bartlettroscoe:
> @rinkug, @bartlettroscoe and @bernhold this is now ready for review.
> 
> The weekly CI check should result in a small list of either bonified broken links or false negatives (e.g. says its broken when its not) to try to check and then either go edit files to fix, add to ignore patterns or just ignore.
>
> ...
> > Consider excluding those URLs. Perhaps with a note (issue) to re-check in a few months.
>
>
> Seems too onerous IMHO since bad certs are a common scenario...we'd be doing this all the time. And, the links still "work" in the browser. So, urlchecker is just being too pedantic about it.

@markcmiller86, IMHO, we should either add ignores for specific URLs like show for the check run [here](https://github.com/betterscientificsoftware/bssw.io/actions/runs/7734371239/job/21088244489) showing:

> Uh oh... The following urls did not pass:
> `/github/workspace/CuratedContent/LanguageReferenceOnLine.md:`
>    ❌️ `https://iitj.ac.in/uploaded_docs/cc/HPC_training/mcmuserguide.pdf`
>     ❌️ `https://www.amd.com/content/dam/amd/en/documents/developer/version-4-1-documents/aocc/aocc-4.1-user-guide.pdf`

or not bother doing an automated check at all.  If this is really a random thing (e.g. due to expired certs)?  IMHO, if we expect this check to fail often, then it will be ignored for a weekly build (because we will expect it to fail).  If there are a lot of false failures, it may not be worth it to automate these checks or we should instead have them part of the PR checks, since that is when people are most likely to look at them (and warn them in the output that failed links may just be due to expired certs).

> Ok, I still need to re-work this a bit. As it stands now, the list of files to check in the non-PR case would be too long. I know what I need to do to fix this.

IMHO, less than 4 min is not too long for a PR check.  Extreme Programming has the [10 minute build](https://agilepainrelief.com/glossary/10-minute-build) as a precondition for CI testing.  Note that it takes longer that long rebuild the preview.bssw.io site (and you can't approve a PR to merge without rebuilding the preview site and manually inspecting the generated article).

A test that is expected to randomly fail often is likely not worth even running.

markcmiller86:
> **IN SUMMARY:** A test that is expected to randomly fail often is likely not worth even running.

So, as I've developed on this, I've been running checkes against links in [this file](https://bssw.io/items/fundamental-resources-for-scientific-computing) with 161 links in it. In the course of the last few weeks of runs, `gnu.org` has gone down and back up, `mpich.org` certs have been found to be invalid (causing a user to have to explicitly permit it to follow the links there), some content on `open-std.org` went behind a password wall, etc. 

Our whole site has ~5500 links (not including `Events` as per reasons stated in the workflow file). The chances that some of the resources supporting those links are not working *during* a check are high. And, I think that is the nature of network-related resources. Hence, @bernhold has suggested that we worry about a link only when it has failed multiple times (e.g. consistently) over the course of a few week's worth of tests.

I can certainly add some logic to keep track of *persistent* failures over the course of several week's worth of runs and then report those in a *special* way.

But, I don't like the idea of having a lot of content with links that regularly go bad. And, that is precisely what will happen if we don't stay on top of this.

bartlettroscoe:
> Hence, @bernhold has suggested that we worry about a link only when it has failed multiple times (e.g. consistently) over the course of a few week's worth of tests.

The problem is that you would need an automated process to keep track of if a link has failed more than once in a row.  An automated process for this would require more infrastructure that we are not likely willing to implement and support.  Otherwise, there is very little chance that people will have the dedication and attention to manually look back at prior link failures to see if there are persistent test failures.

> I can certainly add some logic to keep track of _persistent_ failures over the course of several week's worth of runs and then report those in a _special_ way.

That would be great, but that could take a lot of work to create the infrastructure to get people's attention and make sure that we notice which URL link checks seem to be persistently failing.  This requires a completely different approach to testing.  Instead of monitoring the results of each individual run of `urlchecker`, instead, the "test" is the analysis of a bunch of runs of `urlchecker` over a period of time to look for persistent test failures.  First, GHA does not have any persistence so you would need to use persistent database (DB) to store the results.  Then, you would need a process to pull data from that DB every so often and look for persistent failures.  And then you need a way to alert a group of people about these persistent failures so that some action could be taken to fix them.  Without all of that, these failures will be ignored.

Not that this is the opposite problem that most projects have in trying to detect randomly failing tests in a test suite that should have no random test failures.  For example, Trilinos is working on this problem right now:

* https://github.com/TriBITSPub/TriBITS/issues/600

But in that case, we are looking for flaky tests.  In the case of `urlchecker` the testing process itself is flaky.  (But we have HPC systems that are quite flaky as well, like the ATS2 Sierra system at LLNL.)

> But, I don't like the idea of having a lot of content with links that regularly go bad. And, that is precisely what will happen if we don't stay on top of this.

Agreed, but such a process where you expect lots of random failures is going to ensure that we don't notice the persistent failures when there are so many random test failures, unless you create a automated process as described above.

If you are really serious about this, using CTest to submit to CDash to store test results and then using a Python tool to extract that data from CDash and analyze it is one way to go that I have a lot of experience with.  But this is a good bit of work to set up a robust process for this.  It seems like such a tool and process would need a larger amount of funding and needs to serve many other projects that have similar needs to check for broken URL links (or just persistently failing tests) in a robust way.  (This seems far out of the scope of the PESO project.)

However, this is actually a pretty interesting technical problem and would make a nice research project and publication on its own.  I can't find much existing literature on this case where you expect that any individual test can randomly fail at any time and you need to identify recuring failures.  I actually worked on this problem some as part of the ATDM project and developed at tool (mentioned above) that can make it easy to spot persistent test failures in a sea of randomly failing tests.  But there is more automation that is needed to make that work.

markcmiller86:
[I asked ChatGPT about this too](https://chat.openai.com/share/2a94848f-0037-4e9a-8cb6-9f790bac8c0a) and it seems to suggest many of the same strategies as @bartlettroscoe.

I am roping @vsoch here for comment as she has been developing `urlchecker` and may have thoughts on strategy as well.

markcmiller86:
> And as an alternative to fancy automation, could we post the failed links as a new issue? If the number of bad links is not to large for any one run, we can keep a couple of the issues around and someone can manually compare them for repeats.

Alternatively, we don't do as CI at all (at least not scheduled or maybe even manual...PR checks are still good though). Instead, we use python tools and someone on EB board periodically runs checks and updates list of failure cases and then attempts to *fix* truly broken links as they are encountered.

vsoch:
I'm happy to fix the bugs, but as I mentioned I can't work on this during work time. You just pinged me 1-2 days ago so you haven't really given me a chance to work on it. I would suggest adding the URLs to the skip list for now.

markcmiller86:
> I'm happy to fix the bugs, but as I mentioned I can't work on this during work time. You just pinged me 1-2 days ago so you haven't really given me a chance to work on it. I would suggest adding the URLs to the skip list for now.

@vsoch I mentioned you [here](https://github.com/betterscientificsoftware/bssw.io/pull/1633#issuecomment-1924421278) in this conversation for a different reason...your thoughts and experience, if any, in dealing with and developing a strategy for the larger issue of high probabily of *some* (of the over 5,500 -- and still growing) links likely always failing for reasons other than the link actually needing to be replaced.

vsoch:
Ah gotcha! If you want to make new additions or changes speeder, then I'd recommend changed files: https://github.com/marketplace/actions/changed-files. I use that for container matrices so I only build containers with updated Dockerfile. If you are concerned about existing links breaking (across many files) a dumb thing I do is to segment a list of things (e.g., paths) into equal lists based on matches hashes to calendar month days, then run for the day (a small subset) each night. https://github.com/vsoch/split-list-action. You probably don't want to be checking everything on every PR, every time!

bartlettroscoe:
> I submit that the perfect is the enemy of the good enough. I hope that there are knobs we can tweak to get the failure rate low enough that it is humanly manageable.

Thinking about this some more, I hope that will be the case if we set things up well and have well defined responsibilities (see below)

> The urlchecker action has a number of parameters for retries and timeouts and such.

Retries in the same run of `urlchecker` will not help if a link is broken because of an expired cert or a machine upgrade, etc.  Only running `urlchecker` days later will address that type of issue.

> And as an alternative to fancy automation, could we post the failed links as a new issue? If the number of bad links is not to large for any one run, we can keep a couple of the issues around and someone can manually compare them for repeats. I think the results of the URL check could be posted as an issue using actions.

That will just create a bunch of GH issues that people will ignore because most of these will be false failures .  This will just clutter up the GitHub issue DB.  (I have a lot of experience observing human behavior about reacting to GH Issues.)

You need a process where there are very few false failures.  My experience over 20+ years observing human behavior is that if the number of false failures is above say 30%, most people will just assume it could be a false failure and ignore it. 

> And maybe weekly is too often to run it. How about biweekly or monthly instead, just to reduce the frequency at which someone has to look at the results?
>
> ...
>
>  Instead, we use python tools and someone on EB board periodically runs checks and updates list of failure cases and then attempts to fix truly broken links as they are encountered.

How do people determine what is "truly broken" without manually clicking on the links in a browser?  But I guess if the number of random failures stays low (say less than 10 total on any run of `urlchecker` on every `*.md` file) and you only run this check every month or so, then it would not take too much time to manually check these links.  And my experience is that most external links stay around for quite some time.  It is only older articles with links put in 1+ years ago where you have much of a problem with truly broken links.

So I would suggest something like:

* Set up to run `urlchecker` on all of the *.md files once per month and have it send out emails to the editorial them when it fails (with a link to the GHA output to see what the failures are with some comments in the GHA output about expecting some random link check failures).

* Someone on the editorial team will lock over the failures and manually check all of the broken links (e.g. assuming there less than 10 or so).

But who will be responsible for looking at the link check results?  (If it is everyone's responsibility, then it is no one's responsibility.)

The above seems like a reasonable process, as long as someone is willing to spend 10 minutes every month looking at potential broken links.

> Alternatively, we don't do as CI at all (at least not scheduled or maybe even manual...PR checks are still good though).

It seems that the GHA job will only check links in changed *.md files so I would argue that will not create a lot of false failures (because most *.md files have a few links).  So I would argue we should run this GHA check on all PRs (and add some comments to the GHA output about expecting some random link check failures).

bartlettroscoe:
@vsoch, from the discussion above, I don't think there is anything wrong with `urlchecker` itself or anything that can be done inside of `urlchecker` to address the fundamental problem discussed above.  The problem is that checking for external links and downloading data from those links is inherently going to lead to **some number of random false failures because of issues with the external cites themselves** (like expired certs).  What is needed is a process that runs on top of sampled data produced by a checker like `urlchecker` where you expect some number of random failures where you run it several times over a period of days or weeks to generate sampled datasets.  Then with this set of data collected over a period of time, you have a higher-level analysis tool look over that data for link checks that are consistently failing and then focus only on those.  No one expects a checking tool like `urlchecker` to be able to do that.

markcmiller86:
> No one expects a checking tool like `urlchecker` to be able to do that.

I agree...That said, @vsoch has add a `--no-check-certs` option to the checker to address the expried certs failures. Thanks @vsoch 💪🏻. She is also looking into making `urlchecker` do just `head` requests to make it faster (and not download the linked content as part of the checking process).

I just enlisted @vsoch for her thoughts on this broader question to see if she might be able to suggest solutions.

bartlettroscoe:
>  `@vsoch ` has add a `--no-check-certs` option to the checker to address the expired certs failures. 

@markcmiller86, when is this going to be released in a new version of:

* https://github.com/urlstechie/urlchecker-action

being used by [`check-urls.yml`](https://github.com/betterscientificsoftware/bssw.io/blob/mcm86-check-urls-weekly/.github/workflows/check-urls.yml)?

It looks like there has not been a commit to the `master` branch of that repo in 2 years going back to https://github.com/urlstechie/urlchecker-action/commit/b643b43e2ac605e1475331c7b67247d242b7dce4.

vsoch:
Ping @davidbeckingsale - I have a many ideas for how to address the issues here but I don't have bandwidth soon to work on it, even in free time - there are just other things I need/want to do first. I think an RSE could help here.

vsoch:
@bartlettroscoe that is my library - the action is just a wrapper for urlchecker-python.  And yes, it's been running smoothly for 2 years and hasn't had any feature requests or issues. I'm lead maintainer for these projects, they have about 330 users. Feel free to use something else.

vsoch:
Hi @bartlettroscoe @markcmiller86 I wanted to follow up here - realizing that probably I'm best oriented to work on this, I put aside the work I wanted to do for this afternoon and tackled this issue. I have a new release of urlchecker-python (0.0.35) and a [branch with the action](https://github.com/urlstechie/urlchecker-action/pull/107) that you can test. Importantly:

- the test branch now has support for `--no-check-certs` exposed as no_check_certs under the action
- we use a HEAD request and fall back to GET and close the response (thanks for this insight @markcmiller86 I do believe it will be faster for large pages, which tend to be most these days)
- I fixed (what I consider to be) a bug that we could not target a single file (you now can, if you install / run it locally)

The main issues had nothing to do with the above, and actually came down to changes in the selenium interface. When I debugged I found [this change](https://github.com/SeleniumHQ/selenium/commit/9f5801c82fb3be3d5850707c46c3f8176e3ccd8e) which meant that we were not using a web driver at all and relying solely on requests. The fact that most were failing is a reflection of a huge change in the web - it used to be the case that most would work for requests, and there was only an off case or two where you needed the driver. Now more sites are able to detect (what comes down to) scrapers, and they don't allow this. So since our webdriver was failing, this was resulting in bad results.

@markcmiller86 could you please test out the branch, with the added option to not check certificates, and let me know your feedback? If/when it is good I can merge that PR and do another release of urlchecker-action. Thanks!

markcmiller86:
> @markcmiller86 could you please test out the branch, with the added option to not check certificates, and let me know your feedback? If/when it is good I can merge that PR and do another release of urlchecker-action. Thanks!

@vsoch this is great news! :tada: :muscle:. Thanks so much for tackling this :heart:.

I will try to test before end of weekend.

vsoch:
Looks like we can try to suppress the warnings to clean up the output a bit:

```
import requests
requests.packages.urllib3.disable_warnings() 
```

markcmiller86:
@vsoch I tested as an *action* using the `@update/0.0.35` moniker to specify action version. I tried with and without `no_check_certs`. It seems to be working as expected. I can't speak quantitatively to whether it went any faster but it seems to go a bit faster.

vsoch:
I am preparing a branch with tweaks, such as silencing those warnings. Let me know if you want something else tested before I push a test image and we can try again.

markcmiller86:
> Looks like we can try to suppress the warnings to clean up the output a bit:
> 
> ```
> import requests
> requests.packages.urllib3.disable_warnings() 
> ```

Hmm...I kinda like those. I mean, one issue I have with most actions is that when things go wrong, there is very little information as to why. Maybe tie this output to your `verbose: true` variable as well? I think it could be important to actually see cause if desired.

vsoch:
okay, happy not to do any more work then! :laughing:

vsoch:
I'm going to unsubscribe from notifications here - good luck @markcmiller86 !

bartlettroscoe:
I think we should change this to a manual trigger and merge this.

.github/workflows/check-urls.yml
```diff
@@ -0,0 +1,161 @@
...
+                    continue 2 # ignore this file
+                fi
+                fd=$(echo $f | cut -d'/' -f1)
+                if [ "$ef" = "$fd" ]; then
+                    continue 2 # ignore this dir
+                fi
+            done
+            linkchecker -f utils/LinkChecker/.linkcheckerrc file://$(pwd)/$f >> linkchecker.out || true
+            cat linkchecker.out >> linkchecker-all.out
+        done
```
bartlettroscoe, line 92: Seems like it might be good to break this out into a Python script taking arguments so it can be developed and tested locally.  Also, the Python implementation for some of these operations is a bit cleaner than bash commands.

.github/workflows/check-urls.yml
```diff
@@ -0,0 +1,161 @@
...
+    - name: Process log
+      run: |
+        python utils/LinkChecker/cklcresults.py ${{ github.event_name }}
+
+    - name: Upload artifact
+      if: ${{ github.event_name == 'pull_request' }}
+      uses: actions/upload-artifact@v4
+      with:
+        name: bad-links
+        path: bad_links.txt
```
bartlettroscoe, line 103: Note that I think you might only want to upload the list of bad links for files on the 'main' branch, not a topic branch.  Also, don't you want the full list of *.md files being processed when you generate the list of bad links?

bartlettroscoe: Where is the [`download-artifact`](https://github.com/actions/download-artifact) call to get back a copy of this file?  I guess that still needs to be added.

.github/workflows/check-urls.yml
```diff
@@ -0,0 +1,161 @@
...
+
+    - name: Update link logs
+      if: ${{ github.event_name != 'pull_request' }}
+      run: |
+        git commit -m 'Update link logs'
+        git push origin sched-link-checks
+
+#
+# Keep the recurring failures and definitely bad lists in repo on
+# branch manage-broken-links
```
bartlettroscoe, line 113: This  comment is not correct, is it?  GitHub Action's [job artifacts](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow) system is being used instead, right?

bartlettroscoe (commented):
Just a few comments ...

Wow, this [GitHub job artifacts](https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/storing-and-sharing-data-from-a-workflow) feature looks very useful.

Seems like you should upload and download a JSON file with the list of bad links over say the last 2 weeks and update the counts for the days/times where it was tested to fail.  (A python script that creates and updates that JSON file should not be too bad.)  Then, if a URL link check has filed with every check over the last 2 weeks (or whatever time length), then you report that as a GHA failure for this workflow and print out the bad URL links and their failure history.

bernhold:
A related suggestion:  Along with the reporting, this action could generate a script that would automatically replace each of the bad URLs with text like "(this link is no longer available)".

The idea would *not* be to apply this blindly, but rather for bad links to be investigated and where appropriate alternatives can be found (e.g., relocated or equivalent content), those entries would be changed manually and the lines of the script commented out.   Thus the generated script becomes the final cleanup for the remaining bad links.

The most likely links to go bad are registration and survey links that logically expire or close.  We're likely to get a handful of those each month.  And of course we will not be able to find alternatives for everything else that rots away.

PR #1634, opened 2023-04-18
Skip ci on forks

markcmiller86:
Add `if` logic to CI so we skip CI on forks.

### DIFF
--- a/.github/workflows/merge-pr-to-preview.yml
+++ b/.github/workflows/merge-pr-to-preview.yml
@@ -7,7 +7,9 @@ on:
       - master
 jobs:
   sync-pull-request:
-    if: contains(github.event.pull_request.labels.*.name, 'preview')
+    if: |
+      contains(github.event.pull_request.labels.*.name, 'preview') &&
+      github.repository_owner == 'betterscientificsoftware'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@master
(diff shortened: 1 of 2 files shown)


.github/workflows/merge-pr-to-preview.yml
```diff
@@ -7,7 +7,9 @@ on:
       - master
 jobs:
   sync-pull-request:
-    if: contains(github.event.pull_request.labels.*.name, 'preview')
+    if: |
+      contains(github.event.pull_request.labels.*.name, 'preview') &&
+      github.repository_owner == 'betterscientificsoftware'
```
bartlettroscoe, line 12: If that works, this is fine.  But the way that I implemented this in one of my GitHub repos was to compare `${GITHUB_REPOSITORY}` to the org/repo name (e.g. `TriBITSPub/TriBITS` in my case).

bartlettroscoe (approved):
If you have manually tested this and it works as it should, then go ahead and merge.  Otherwise, it looks reasonable but I don't have any personal experience with this property `github.repository_owner`.

markcmiller86:
> If you have manually tested this and it works as it should, then go ahead and merge.

Yup. Tried on my own fork. You can see (most recent) cases [here](https://github.com/markcmiller86/bssw.io/actions/workflows/merge-master-to-preview.yml) where it was successfully skipped.

PR #1815, opened 2023-09-13
Additional master --> main changes needed

bernhold:
@bartlettroscoe I believe these changes are also necessary or appropriate.  Most of them are references to our image folder.  A few are places where "master" has been used in an article and IMHO it can easily be changed to something else.


bartlettroscoe (approved):
@bernhold, good catch!  My regex must have missed the pattern:

```
https://github.com/betterscientificsoftware/bssw.io/raw/master/images/<some-file>
```

somehow.

Also, I did not update the content of the existing articles to replace 'master' with 'main' but those changes you made look good.

Issue #1864, opened 2023-11-13
Add support for codespell checking


bartlettroscoe:
@betterscientificsoftware/bssw-editorial-board 

**Related Issues:**

* #334

## Description

As I mentioned in https://github.com/betterscientificsoftware/bssw.io/issues/334#issuecomment-1491903088, I have some experience with the tool codespell.  It checks for commonly misspelled words and is pretty easy to work with.  I am looking at setting it up for the `bssw.io` Git repo with the config file `.codespellrc`

```
[codespell]
check-filenames =
check-hidden =
# Disable warnings about binary files
quiet-level = 2
skip = */.git,*/*.pdf
#ignore-words-list =
```

and it seems to be showing a lot of misspellings that we have missed shown below (as of the commit 392d0c3b):

<details>

<summary><b>codespell run on bssw.io repo as of commit 392d0c3b</b> (click to expand)</summary>

```
Used config files:
    1: .codespellrc
./.github/PULL_REQUEST_TEMPLATE.md:5: recieved ==> received
./.github/ISSUE_TEMPLATE/blog-article.md:11: recieved ==> received
./.github/ISSUE_TEMPLATE/event.md:13: refered ==> referred
./.github/ISSUE_TEMPLATE/other.md:10: recieved ==> received
./Articles/Blog/10-new-hacks.md:4: Nam ==> Name
./Articles/Blog/10-new-hacks.md:4: Nam ==> Name
./Articles/Blog/2018-11-messer-porting.md:45: programing ==> programming
./Articles/Blog/2020-03-rate-your-project_UNPUB.md:2: incase ==> in case
./Articles/Blog/2020-03-rate-your-project_UNPUB.md:6: somes ==> some, sums
./Articles/Blog/2020-11-PSIP4HDF5.md:14: aways ==> always, away
./Articles/Blog/2020-12-2021BSSwFellowshipAnnouncement.md:3: Nam ==> Name
./Articles/Blog/2020-12-2021BSSwFellowshipAnnouncement.md:3: Nam ==> Name
./Articles/Blog/2020-12-2021BSSwFellowshipAnnouncement.md:104: Nam ==> Name
./Articles/Blog/2021-03-2021BSSwFellowshipNSFAnnouncement.md:3: Nam ==> Name
./Articles/Blog/2021-03-2021BSSwFellowshipNSFAnnouncement.md:3: Nam ==> Name
[...]

PR #1873, opened 2023-11-22
Add support for codespell (#1864)

bartlettroscoe:
@betterscientificsoftware/bssw-editorial-board 

Addresses:

* #1864

I fixed a bunch of misspellings and typos found by codespell.

Look at how few words I had to list on 'ignore-words-list' in the .codespellrc file.

This tool is much easier to use and requires less work that adopting a standard spell checking tool.


bernhold (approved):
Looks great! Thanks for doing this!

PR #1874, opened 2023-11-22
Add GitHub Action using codespell to check files and commit msgs (#1864)

bartlettroscoe:
Addresses:

* #1864

I copied this from the TriBITS repo which was provided by Kyle Edwards from Kitware (this is what they use for CMake).

### DIFF
--- a/.codespellrc
+++ b/.codespellrc
@@ -4,5 +4,5 @@ check-hidden =
 # Disable warnings about binary files
 quiet-level = 2
 # Ignore Words List: Formal names, acronyms, other words
-ignore-words-list = nam,hass,hart,wil,damon,leary,cruzes,nwo,idaes,precice,rcall,ags,coo,theses,nd,algebraical,unsecure,commun,ore,re-use
+ignore-words-list = nam,hass,hart,wil,damon,leary,cruzes,nwo,idaes,precice,rcall,ags,coo,theses,nd,ba,algebraical,unsecure,commun,ore,re-use
 skip = */.git,*/*.pdf,*/_config.yml,*/jquery.shuffle.min.js,*/build_tests/*,*/utils/tests/*.md
--- a/.github/workflows/codespell.yml
+++ b/.github/workflows/codespell.yml
@@ -1,0 +1,66 @@
+name: BSSw.io Codespell
+
+on:
+  pull_request:
+    branches: [ main ]
+
+jobs:
+  codespell-source:
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v3
+
+      - name: Print GitHub Actions Environment Variables
+        run: |
+          set | grep "^GITHUB_"
+
+      - name: Update OS
+        run: |
+          sudo apt-get update -y
+      - name: Install packages
+        run: |
+          sudo apt-get install -y codespell
+
+      - name: Run Codespell on source code
+        run: |
+          codespell
+
+  codespell-commit-messages:
+    runs-on: ubuntu-latest
+
+    if: github.event.pull_request != null
+
+    steps:
+      - name: PR commits + 1
+        run: echo "PR_FETCH_DEPTH=$(( ${{ github.event.pull_request.commits }} + 1 ))" >> "${GITHUB_ENV}"
+
+      - name: Checkout PR branch
+        uses: actions/checkout@v3
+        with:
+          ref: ${{ github.event.pull_request.head.sha }}
+          fetch-depth: ${{ env.PR_FETCH_DEPTH }}
+
+      - name: Fetch base ref
+        run: |
+          git fetch origin ${{ github.event.pull_request.base.sha }}
+
+      - name: Print GitHub Actions Environment Variables
+        run: |
+          set | grep "^GITHUB_"
+
+      - name: Update OS
+        run: |
+          sudo apt-get update -y
+      - name: Install packages
+        run: |
+          sudo apt-get install -y codespell
+
+      - name: Run Codespell on commit messages
+        run: |
+          result=0
+          for commit in $(git rev-list HEAD ^${{ github.event.pull_request.base.sha }}); do
+            echo "Running codespell on commit message of $commit..."
+            git show --format=%B -s "$commit" | codespell - || result=1
+          done
+          exit $result
--- a/Articles/Blog/2023-02-anatomy-rse-team.md
+++ b/Articles/Blog/2023-02-anatomy-rse-team.md
@@ -145,7 +145,7 @@ post](https://bielsnohr.github.io/2022/10/14/pasc22-minisymposium-summary.html#f
 
 ### Author bio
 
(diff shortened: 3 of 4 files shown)


bartlettroscoe:
Hello @bernhold, this adds GHA checks for the files and the commit messages.  I can post local Git hooks for people that want these checks run in their local repos when doing work.  (I have these set up for myself.)

bartlettroscoe:
NOTE: You can see how this GHA runs when it passes [here](https://github.com/betterscientificsoftware/bssw.io/actions/runs/6959328783/job/18936316637) and when it fails [here](https://github.com/betterscientificsoftware/bssw.io/actions/runs/6959238409).

bartlettroscoe:
@bernhold, @betterscientificsoftware/bssw-editorial-board 

NOTE: We may consider removing the check of the commit messages because I don't think most developers know how to fix commit messages for existing commits (and I don't think the GitHub interface supports that).

bernhold:
@bartlettroscoe is there any mechanism to make spelling exceptions on a per-file basis?

The use case that I've encountered is a blog article that's using formatting to highlight the first letters of words to make an acronym.  So we see things like `**S**poofing`.  In cases like this, the "poofing" would presumably show up as a spelling error.  But really, it is a special case which is likely unique to this file.  Anywhere else, "poofing" is still going to be an error.  So it would be better if there were some way to make the exception file-specific.

If such a mechanism exists, the next question is how easily can we incorporate it into our setup?

Thanks

bartlettroscoe:
> @bartlettroscoe is there any mechanism to make spelling exceptions on a per-file basis?

Unfortunately, no.  See:

* https://github.com/codespell-project/codespell/issues/1212

I don't know of a better solution than to add `poofing` to the global ignore-word-list in:

https://github.com/betterscientificsoftware/bssw.io/blob/054dd1b12749eb9507d6abddba1d5e9ea87c5345/.codespellrc#L6-L7

By doing that, you ignore all usages of `poofing` like misspelling `proofing`, so that is unfortunate. 

But I would argue that it is better to skip a few misspellings like this than to skip all misspellings.  Just look how many misspelling were caught and fixed in:

*  #1874 

I think this is a case where we should not let the perfect be the enemy of the better.  That is, using codespell with a lot of (unfortunate) global ignores is better than not using it at all.  IMHO, even with this problem, codespell (which uses a dissallow-list approach) is less work to maintain than a standard spell checker that uses a allow-list approach.

I would suggest that we go ahead and merge this PR and see how this goes.  If we have to disable it or do something else, then we can do that.  But I think having this in place will help some.

We might add some documentation somewhere for how to exclude words and entire files and directories perhaps.  But if you know about the existence of the file `~/.codespellrc`, then that should be pretty self-explanatory, hopefully.

markcmiller86:
> > @bartlettroscoe is there any mechanism to make spelling exceptions on a per-file basis?
> 
> Unfortunately, no. See:
> 
> * [Support inline ignores codespell-project/codespell#1212](https://github.com/codespell-project/codespell/issues/1212)

@bartlettroscoe can't you maybe fake it with some additional scripting...

* allow an xml comment in each file of the form
   ```
      <!-- Codespell Ignores: poofing, requirements, foobar -->
   ```
* then, *before* codespell is invoked on a file, we have a script that grabs this from the file (if it exists), creates a temporary (*global*) ignore list and then processes the file.

bernhold:
Is there some other spell checker that would provide this file-specific capability?  Is there some reason to choose codespell over others that might have this feature?

bartlettroscoe:
Codespell seems to be widely used and is easy to install on various platforms.  I don't know of a competitor for Codespell for the disallow-list appraoch.

markcmiller86:
> For example, if the driver found a file called `<filename>.codespell-ignore-words-list`, then it would use the ignore list in that file.

I think having to put words to ignore in a wholly separate, per-content-file file, makes things even more challenging. It might be ok to have a *single* global (json or yaml) file that lists each content file and the ignore words for that file...

`.codespell-ignore-lists`

```
- ./Blogs/foo.md
   - poofing
   - foobar
- ./CuratedContent/bar.md
   - gorfo
```

markcmiller86:
Honestly, my first thought would be as I originally propsed...an xml comment (maybe at the top or bottom of the file) that is parsed out of the file and used to produce an invokation of `codespell`.

bartlettroscoe:
> Honestly, my first thought would be as I originally propsed...an xml comment (maybe at the top or bottom of the file) that is parsed out of the file and used to produce an invokation of `codespell`.

Yea, that would work too.  In fact you could put it in any XML comment block like:

```
<!---
...
codespell-ignore-words-list: <w1>,<w2>,...
...
--->
```

and then you would first need to run a tool that read the file `<filename>` looking for `codespell-ignore-words-list:` and extract out `<w1>,<w2>,...` if it was found like:

```
ignore_words_list=$(extract_codespell_ignore_words_list.py <filename>)
if [[ "${ignore_words_list}" != "" ]] ; then
  codespell --ignore-words-list=${ignore_words_list} <filename>
else
  codespell <filename>
endif
```

The only other complication is that you would need to read `skip` in:

https://github.com/betterscientificsoftware/bssw.io/blob/054dd1b12749eb9507d6abddba1d5e9ea87c5345/.codespellrc#L8

and apply that yourself in the outer driver script to downselect what files you pass to codespell.

I can understand why the codespell developers did not want to take on the job of figuring out how to allow meta-data for codespell stored in the files themselves for every different type of file format.  That is a lot of extra responsibility to take on.

markcmiller86:
> I can understand why the codespell developers did not want to take on the job of figuring out how to allow meta-data for codespell stored in the files themselves for every different type of file format. That is a lot of extra responsibility to take on.

Yes and no. I mean, `codespell` does operate on ascii files only, right? So, all they really need to do is specify a *magic ascii phrase* of some kind (unlikely to ever appear in any real ascii content) that can be used as a sort of meta-escape character so that when it is encountered, data following it (or perhaps bracketed by it, etc.) can then be passed to codespell. That doesn't rely on any knowledge of the particular ascii format in use. It just requires some ingenuity and creativity...

```
@codespell-ignore-words{...}
```

Then, anyone wanting to use that needs to find *their own way* to *harmlessly* embed it in their ascii files...

* xml: `<!-- @codespell-ignore-words{...} -->`
* C++: `// @codespell-ignore-words{...}`
* rst:
   ```
     ..
       _@codespell-ignore-words{...}
   ```
* yaml, makefile, python, etc: `# @codespell-ignore-words{...}`
* markdown: `[//]: # @codespell-ignore-words{...}`
* fortran: `c      @codespell-ignore-words{...}`
* etc.

bartlettroscoe:
> Then, anyone wanting to use that needs to find their own way to harmlessly embed it in their ascii files...

Right, but what if you want to provide text documenting the usage `codespell-ignore-words`, like shown above?  To be robust, codespell needs to know how comments are handled in different file types.  (Note that emacs/ispell knows how comments are used for various file types.)

bartlettroscoe:
> Is there some other spell checker that would provide this file-specific capability? Is there some reason to choose codespell over others that might have this feature?

@bernhold, see https://github.com/betterscientificsoftware/bssw.io/issues/1864#issuecomment-1843994157.

bartlettroscoe:
Okay, @bernhold signed off above and so did @markcmiller86 in https://github.com/betterscientificsoftware/bssw.io/issues/1864#issuecomment-1844473302 so I will merge this PR and see how this goes.

PR #1885, opened 2023-11-30
Add Track to the Short Articles

pagrubel:
Added track to each short article using the format:
Track: <Track Name>

I can correct it quickly if the format is wrong.


rinkug:
LGTM

PR #1928, opened 2023-12-16
Add basic footnote support to wikize_refs

### LINKED ISSUE #1122: Enhance wikize_refs.py to support ordinary footnotes
In #1121 there is [this comment](https://github.com/betterscientificsoftware/bssw.io/pull/1121#issuecomment-955765272).

The document is trying to use a `wikize_refs.py` style of references to do ordinary footnoting.

With the reference links adjusted as per [this comment](https://github.com/betterscientificsoftware/bssw.io/pull/1121#issuecomment-955766497) it almost works. Two issues are...

* It still produces (invalid) off page links for the footnote text itself
* It places the footnotes in a section titled `References` and would co-mingle them with any actual references

I think both issues are easily fixed by detecting any link definitions whose url-field is `#`. It that case, there exists real footnotes in the document and then we keep those in a separate list and output them at the bottom of the document *ahead* of the `References` section, if it also exists.

markcmiller86:
Resolves #1122 

- Adds ability to handle *basic* footnotes.
- Renders basic footnotes, if present, in their own block ahead of the **References** section.
- Adds tests for this mode also.
- Adds `-u` (`--up-to-date`) option which ensures exit code is non-zero if `wikize_refs.py -i` would change the file.
- Removes *two-column* mode, which never really did what we wanted to begin with.


rinkug:
@markcmiller86 : I dont know enough to review this. If you think its good enough, then please merge it

Issue #1469, opened 2022-10-14
Long-term goal: meaningful preview without having to go to the preview site


bernhold:
I find that that the thing that makes article development really tedious is having to rebuild the preview site repeatedly.  You have to move to a different tool (the preview site), fiddle, and wait while it rebuilds.  Rebuilds are currently taking O(10 min).  It can really slow down iterations on getting the format of an article right.

One approach to addressing that is to align the way BSSw styles things closer to the way popular tools like GitHub and web browsers with default CSS settings style content. That's captured in #706.

Another approach would be to provide a way to preview a single page that retains the most important aspects of what the front-end is doing for us and uses the same CSS as the site.  Maybe a very simple way to render the markdown to HTML which can then be loaded locally into a browser and reference the bssw.io CSS files or something.  It would need to preserve the key elements of the processing the front-end currently does to articles.  It seems like one way to approach that might be to try to reduce the special stuff the front-end is doing that materially changes the content of the document.  I think if we captured the right 80-90% of the structure and formatting of the document, it could reduce the need for previews through the preview site significantly.

If #706 is, in essence, making our CSS closer to "standard", then this idea might reducing the "hidden" transformations that the front-end applies to the documents it ingests that isn't typical processing markdown documents.

bernhold:
I remember talking with Parallactic about this some time ago and they suggested that they could probably add a feature to the front-end that would allow us to point (somehow) to a particular file and have it rendered (as best it can without having all of the database information, etc.).

Issue #706, opened 2021-02-11
Can we make it easier for contributors to view content as it would be rendered on the site?


bernhold:
At present, the renderings of markdown content available to contributors from GitHub's UI and other easily authoring tools can sometimes be significantly different from what they'll see on the preview (or production) site.  It would be easier on contributors and our editors if they were closer. (At least specific markdown elements should be rendered similarly.)

At present, we're pretty sure that the front-end uses a different flavor of Markdown than GitHub, and we know that the CSS for the site is rather different than the default that most authoring tools us.  It would be nice to systematically move them closer together.

bernhold:
A challenge with changing the front-end's Markdown processor or CSS is that we already have a lot of content using what's currently there.  We'd need to make sure old content transitions smoothly, which could be challenging.

An alternative might be to add a metadata tag to all existing content indicating that it uses "old rendering" and apply different CSS (maybe even different markdown processors?) depending on the metadata setting.

bernhold:
Ross encountered problems with greater and less-than symbols (`<>`) being stripped out, even in blocks that are supposed to be taken verbatim.  See here: https://github.com/betterscientificsoftware/bssw.io/pull/1156#issuecomment-999063412

PR #2141, opened 2024-08-23
Fixes for How to Improve Testing for CSE Software article

bartlettroscoe:
See commits for details.

### DIFF
--- a/Articles/Blog/HowtoImproveTestinginyourCSESoftwareProject.md
+++ b/Articles/Blog/HowtoImproveTestinginyourCSESoftwareProject.md
@@ -88,7 +88,17 @@ Topics: Testing
 Track: how to
 --->
 
-[1]: https://bssw.io/items/working-effectively-with-legacy-code "Working Effectively with Legacy Code {Michael Feathers. Prentice Hall, 2005}"
-[2]: https://www.oreilly.com/library/view/test-driven-development/0321146530/ "Test Driven Development {Kent Beck, Addison-Wesley Professional, 2003, ISBN: 0321146530}"
-[3]: https://bssw.io/items/code-complete-a-practical-handbook-of-software-construction "Code Complete: Second Edition}"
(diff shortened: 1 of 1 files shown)


bartlettroscoe:
Here is the generated article:

* https://preview.bssw.io/blog_posts/how-to-improve-testing-for-cse-software

The multi-level numbering is unfortunate in that it does not change the numbering style for the nested lists.  For example, GitHub formats this with 1, 2, 3 ... for the first level and i, ii, iii, ... for the second level as:

image

but the bssw.io site generator just uses 1, 2, 3, .... for every level like:

image

That is unfortunate.  This does not look very good.

It seems the alterative is to use manual HTML list formatting as described at:

* https://www.reddit.com/r/gitlab/comments/14tajex/markdown_for_ordered_lists_with_letters_and/

Is it worthwhile trying that?

But this still looks **way** better than the current production article at:

* https://bssw.io/blog_posts/how-to-improve-testing-for-cse-software

which has formatting completely messed up, for example showing:

image

bartlettroscoe:
So, I created the commit https://github.com/betterscientificsoftware/bssw.io/pull/2141/commits/79c2a2125b6e2edb80b719f0e6cb00b78837be70 that tries to take over the formatting with HTML and the bssw.io site generator ignores the `<ul type="a">` attribute and just shows the same formatting:

image

So that was a waste of time.  But at least we now know that resistance is futile and we can just acquiesce the to borg bssw.io site generator and live with what we are given😊 (But it would be nice to address #1469 and #706.)

bartlettroscoe:
This is ready to merge.  Since this is a big improvement over the current production article, I will just merge it and rebuild the production site.

bartlettroscoe:
We see the updated article deployed at:

* https://bssw.io/blog_posts/how-to-improve-testing-for-cse-software

and it looks quite a lot better, IMHO.

Issue #2248, opened 2025-02-06
Possible Blog Article: Best Practices for Multi-Project Continuous Integration


ryanmrichard:
Similar to #2242, I am looking to write a blog post summarizing my 2024 BSSw Fellowship project. The article will have the same title as this issue and will be designed to summarize the results collected on the [Multi-Project DevOps Website](https://multiprojectdevops.github.io) I created to accompany the project.

Like #2242, this issue is advanced notice for the blog editors to make sure they approve the topic and are willing to receive the article.

bernhold:
Hi @ryanmrichard, please go ahead, we're looking forward to your contribution!

As to the logistics, I think this is your first blog for us, right?  The essentials are similar to events, but the expected content is a little different.  My suggestion would be to find a published blog article that looks roughly like how you want to structure your article and use that as a template.  But we're always happy to work with you if you have questions along the way.

Can we set a target date for you to deliver a draft?  That way we can start planning for it.  Thanks

ryanmrichard:
@bernhold I've actually got a draft already. I'll open a PR in a second.

bernhold:
Article has been published

PR #2249, opened 2025-02-07
Article on best practices in multi-project CI/CD

ryanmrichard:
This PR is the article mentioned in #2248.


bernhold:
@ryanmrichard, you seem to be using a number of tinyurl links in the article.  Is there a particular reason for that?  I would generally prefer to expose the full direct link instead of a shortened reference.  It allows readers to use the hover preview of the link to get useful information and decide whether they want to follow it or not.  It is generally considered more friendly from a cybersecurity standpoint too.

Unless there's a strong reason, I'm going to replace the tinyurls with their expansions.  Please let me know ASAP.  Thanks

ryanmrichard:
@bernhold I used the tinyurl links as a force of habit. In my other projects I'm obsessive about not going over 80 characters per line and markdown doesn't have a great way for splitting long URLs (at least that I've found). I think your changes look good!

PR #2251, opened 2025-02-14
Create ExperiencesIntegratingMOABandVisIt.md

markcmiller86:
@iulian787, @vijaysm with help from @markcmiller86 are drafting an article about the integration of MOAB with VisIt. We are using this **DRAFT** PR to develop the content.


markcmiller86:
@vijaysm and @iulian787, you may wind up having to fork from here and submit PR's into this PR  to contribute your pieces. If this proves to be too onerous, please let us know. We can adjust workflow to accomodate.

vijaysm:
@markcmiller86 No, that workflow sounds good to me.

markcmiller86:
@vijaysm and @iulian787 I have started writing and will plod along here periodically. My practice is typically to write often a lot more than necessary and then pare down to get to the final product. So, don't worry too much about length at this point. We can adjust to ensure we emphasize the desired items and don't over saturate with irrelevant details.

markcmiller86:
@vijaysm and @iulian787 just checking in. We should probably try to get this finalized.

markcmiller86:
> Agreed! Are you back from your break, @markcmiller86 ?

Yes. If we need to meet virtually too, lemme know and we can try to re-engage on that too.

vijaysm:
Hi @markcmiller86 , I'm open to moving this elsewhere, with a goal/target in mind for publishing the article in a specific venue. I'd suggest JOSS or JORS if we had a dedicated repository for the plugin itself that can be built against VisIt. But since we are more tightly integrated with VisIt, I'm unsure what other venue would be appropriate.

bernhold:
Hi @markcmiller86, @vijaysm, and @iulian787, thanks for inviting us to take a look at it.

From the length perspective, I'm not yet concerned, though I notice comments above that suggest there may still be a fair amount to come. If you thought the length would need to more than double to complete the article, then I'd start to question the length.

But perhaps the more significant question in my mind is who is the real audience for this article, and what do you want them to take away from it?  What I read so far is a fairly technical discussion of the mechanics involved in creating a new VisIt plugin.  But I do not yet see what broader takeaways there might be for scientific software developers looking for ideas to help them make their software better.  That leads me to question, based on what you've written so far, whether BSSw.io is, indeed, the right audience for this?  If you're looking to document the plugin, how you came to the implementation, and things like that, without necessarily having a message about "better scientific software", then I think another venue might be more appropriate.  Maybe things like JOSS or JORS as @vijaysm suggested, or perhaps there is some more traditional journal in your space that focuses more strongly on the software/implementation side of things (in other domains I'd point to something like Computer Physics Communications or maybe ACM TOMS).  Depending on your goals, perhaps even a technical report or an arXiv paper would be useful?

I hope this makes sense. Happy to talk in more detail if it would help.

markcmiller86:
Thanks @bernhold for taking a look and sharing thoughts. I agree with all your remarks.

So, we'll move this somewhere else and target a different publication.

That said, I do think there is a compelling story here that has to do with LLNL and ANL software teams working together and learning each other's data models and parallel execution paradigms well enough to achieve a useful integration of MOAB and VisIt.

markcmiller86:
@vijaysm and @iulian787, I have moved this article to a *private* repo in my personal space, here, https://github.com/markcmiller86/bssw.io/blob/mcm86-04sep25-moab-visit-integration/Articles/Blog/ExperiencesIntegratingMOABAndVisIt.md

I have invited you both to the repo. So, you should be able to edit in place. No forking needed.

markcmiller86:
Whoops...wait a minute. I've made a mistake in repo paths, etc. Will fix momentarily.

markcmiller86:
Ok, sorry for bad link in previous comment. Go here, https://github.com/markcmiller86/markdown-collab/blob/main/ExperiencesIntegratingMOABAndVisIt.md.

PR #2225, opened 2025-01-14
Ethical source licenses

markcmiller86:
# Description

Draft article.

I drafted this because I had put together an IM on this topic and thought it would be easy to capture that as a CC too. Unfort, what I've drafted is a bit askew of a CC but I am not sure I have knowledge to try to develop further into a blog article.

EB Member: @rinkug 

Resolves #`<issue-id>`

`<Other minimal information about the PR.>`

## PR checklist for files displayed on bssw.io site

* [x] ***[Author]*** `@mention` the BSSw.io editorial board member `@<eb-member-id>` in **Description** above assigned to shepherd your PR.
* [ ] ***[Author]*** Add the `<issue-id>` in the **Description** above for the associated GitHub Issue.
* ~~[ ] ***[Author]*** Ensure `wikize_refs.py -i <base>.md` is run and commit (if using [wikize_refs.py])~~.
* [x] ***[Author]*** Inspect the content in the `*.md` file(s) as rendered in GitHub for this PR.
* [x] ***[EB Mem]*** Assign this PR to the EB member `<eb-member-id>`.
* [x] ***[EB Mem]*** Assign this PR to the author of the PR `<pr-author-id>`.
* [x] ***[EB Mem]*** Add one or more Reviewers.
* [x] ***[EB Mem]*** Add label `content: <content-type>` for the type of contribution.
* [x] ***[EB Mem]*** Add to Project `Content Development` (see [Content Development]).
* [x] ***[EB Mem]*** Add [meta-data] to the `*.md` file(s) (set `Publish: yes`).
* [x] ***[EB Mem]*** Add label `preview` (so PR branch will be merged to 'preview' branch and watch for possible merge failures).
* [x] ***[EB Mem]*** Rebuild [preview] site and confirm new content is there, renders correctly and is returned in searches.
* [x] ***[Author]*** Make any final changes to the PR based on feedback.
* ~~[ ] ***[Author]*** Ensure `wikize_refs.py -i <base>.md` is run and commit (if using [wikize_refs.py]).~~
* [x] ***[EB Mem]*** Rebuild [preview] site and re-confirm content looks correct.
* [x] ***[EB Mem]*** Ensure at least one reviewer signs off on the final changes.
* [x] ***[EB Mem]*** Move the PR to "Ready to Publish" in [Content Development].
* [x] ***[EB Mem]*** Leave comment and @mention `@betterscientificsoftware/bssw-maint` (**BSSw Maint**) asking to carry out final publication steps.
[...]

PR #2260, opened 2025-02-26
Suggestions for EthicalOpenSource.md

rinkug:
Scientific example for  #2225 2225
* Recommended a more science-related example
* Recommended neutral ending
* suggesting licensing as a topic for this cc :)

### DIFF
--- a/CuratedContent/EthicalOpenSource.md
+++ b/CuratedContent/EthicalOpenSource.md
@@ -5,7 +5,7 @@ Is your open source code being used for good or evil?
 <!--- deck text end --->
 
 #### Contributed by [Mark C. Miller](https://github.com/markcmiller86)
-#### Publication date: February, 15, 2025
+#### Publication date: March 15, 2025
 
 Resource information | Details
 :--- | :---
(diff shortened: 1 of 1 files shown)


markcmiller86 (commented):
I wanna decline this change. It winds up using `considering` and `considerations` in the same sentence which is repetitive. I am fine with removing `real-world` though.

markcmiller86 (approved):

CuratedContent/EthicalOpenSource.md
```diff
@@ -27,23 +27,16 @@ Are developers of the software ok if it is used, for example, in creation of dee
...
 By contrast, of the approximately 8,500 packages in [Spack](https://spack.io), only one uses an ethical-source license.
 
-To put these ethical-source licenses in perspective, it is worth considering if and how ethical use restrictions have been applied to other real-world products or services.
-A good example is chemicals.
-Chemical manufacturers all over the world face various restrictions in what they produce and to whom and where they can distribute/sell it.
-But, these restrictions are rarely if ever based on *ethical* considerations and are instead based on things like environmental or human health impacts often supported by a volume of scientific evidence.
+To put ethical-source licenses in perspective, it is worth considering how ethical use considerations have been applied to other technologies, such as cryptographic algorithms.
```
markcmiller86, line 30: Apologies but this example does not work for me.

I am using an analogy here between licensing of open source software (which is a form of controlling how said software is manufactured and distributed) and how some other examples of tangible (I used `real-word`) items people can easily understand are controlled. Chemical manufacturing is a great example IMHO, because raw chemicals (much like a lot of software infrastructure) are incorporated into many final products that have both positive and negative impacts on human life...and which have the potential to be used for good or for evil.

I didn't want to end nuetrally because I myself do not hold a neutral opinion regarding ethical use licensing nor should, IMHO, bssw. However, if there is broad disagreement with that, I am also fine not moving this item foward as is to publication.

But, unfortunately, I am not fine with these changes.

markcmiller86: Maybe a better example, if we really must have a software example, is social media. I think there is plenty of evidence that social media has the potential for harm as well as good.

OTOH, a lot of countries are indeed moving to pass laws to restrict social media in specific ways which is not at all consistent with open source licensing in a couple of important ways. First, open source licenses are a construct of the community itself, trying to adopt practices/laws that govern manufacture and distribution of its product, software. Second, the rational is not so much that the software itself is problematic but that certain *uses* of the software are problematic.

rinkug:
Closing as this is invalid

PR #2500, opened 2026-03-24
Fix wikize_ref.py and add Agent Skill

bartlettroscoe:
Used codex to fix a defect with wikize_refs.py and added an agent Skils.md file so agents can use wikize_refs.py easier.


bartlettroscoe:
No harm in having this so I will go ahead and merge this.

PR #2548, opened 2026-05-29
some changes for @bernhold

markcmiller86:
@bartlettroscoe I think this fixes `wikize_refs.py` for @bernhold. But, can @bernhold confirm it produces desired behavior.

### DIFF
--- a/docs/pages/bssw/bssw_wikize_refs.md
+++ b/docs/pages/bssw/bssw_wikize_refs.md
@@ -44,7 +44,7 @@ at the end of the markdown file, of the form
   - **Note**: The post-processing script renumbers them 1...N
 - `URL`: (**REQUIRED**) The URL for the reference
 - `TIT`: (*optional*) A (short) title for the reference (will also appear as tool-tip during hover)
-- `{BIB}`: (*optional*) Remaining full bibliographic information for the reference including the full title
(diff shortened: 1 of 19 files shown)


markcmiller86:
So, you should understand that the requested changes are backwards compatibility breaking...that is if you regen an existing file, you will get diffs due to changes in heading level (now `##` instead of `###`) and removal of `SHORT TITLE<br>` from entry added in reference table at end.

markcmiller86:
We could add an argument (e.g. `--legacy-mode`), I guess, that tells it to behave the *old* way when that may be needed.

bernhold:
> So, you should understand that the requested changes are backwards compatibility breaking...that is if you regen an existing file, you will get diffs due to changes in heading level (now `##` instead of `###`) 

> We could add an argument (e.g. --legacy-mode), I guess, that tells it to behave the old way when that may be needed.

Actually, since you retired, we've modified the CSS and the article markup so that we can consistently go down the hierarchy of headings instead of skipping from h1 to h3.  That's actually almost entirely down to Sandbox, to give credit where it is due.  So I don't think backward compatibility is a serious concern.  Personally, I don't feel that we need a flag to handle this.

Also, we almost never need to rerun wikize-refs once we get the article published.  Maybe we'll catch something just afterwards, but not after any time has elapsed.

PR #2582, opened 2026-08-25
Pearc26 blog post

adamlavely:
This is a blog post about PEARC26.


bernhold:
Hi @adamlavely , what do we need to move this from draft to ready to publish?  Anything I can help with?  I'd like to publish this in our early September slot, which is about three weeks from now.  Thanks

PR #2576, opened 2026-08-13
BSSwF 2027 applications open

gonsie:
@curfman @bernhold 

Please do NOT merge until we are confirmed for 2027 class and applications have been marked as *open* in Linklings system.

PR #2590, opened 2026-09-01
update time/date of webinar in timeline

adamlavely:
The timeline still had the old time/date for the webinar


bernhold:
Subsumed into #2576

